PR #162, opened 2015-11-18
make get_latlng return tuple instead of None

wseo88:


apps/geolocations/utils.py
```diff
@@ -20,7 +20,7 @@ def get_latlng(location_name):
         latlng = get_latlng_google(location_name)
         if latlng is None:
             # an exception occurred; possibly hit Google API limit
-            pass
+            latlng = (None, None)
```
jontsai: @wseo88 I like the convention `(None, None,)`

Please add the comma after the last element and I'll approve and merge

PR #270, opened 2018-07-16
Adds default autoplay argument and handles youtube

ojameso:
Autoplay defaults to false, so it should work with existing code.

### DIFF
--- a/lib/oembed/utils.py
+++ b/lib/oembed/utils.py
@@ -7,12 +7,17 @@ from htk.lib.oembed.cachekeys import OembedResponseCache
 from htk.lib.oembed.constants import *
 from htk.utils.request import get_current_request
 
-def get_oembed_html(url):
+def get_oembed_html(url, autoplay=False):
     """Gets the oEmbed HTML for a URL, if it is an oEmbed type
     """
     oembed_type = get_oembed_type(url)
+
     if oembed_type:
-        html = get_oembed_html_for_service(url, oembed_type)
+        if oembed_type == 'youtube':
+            autoplay = autoplay
+            html = youtube_oembed(url, autoplay)
+        else:
+            html = get_oembed_html_for_service(url, oembed_type)
     else:
         html = None
     return html
--- a/lib/oembed/utils.py
+++ b/lib/oembed/utils.py
@@ -14,8 +14,7 @@ def get_oembed_html(url, autoplay=False):
 
     if oembed_type:
         if oembed_type == 'youtube':
-            autoplay = autoplay
-            html = youtube_oembed(url, autoplay)
+            html = youtube_oembed(url, autoplay=autoplay)
         else:
             html = get_oembed_html_for_service(url, oembed_type)
     else:


lib/oembed/utils.py
```diff
@@ -7,12 +7,16 @@
...
 
-def get_oembed_html(url):
+def get_oembed_html(url, autoplay=False):
     """Gets the oEmbed HTML for a URL, if it is an oEmbed type
     """
     oembed_type = get_oembed_type(url)
+
     if oembed_type:
-        html = get_oembed_html_for_service(url, oembed_type)
+        if oembed_type == "youtube":
```
ojameso: @jontsai Do you prefer ternary operator here? It would be a long like.
`html = youtube_oembed(url, autoplay) if oembed_type == 'youtube' else get_oembed_html_for_service(url, oembed_type)`

jontsai: single quotes

lib/oembed/utils.py
```diff
@@ -7,12 +7,16 @@
...
+            html = youtube_oembed(url, autoplay)
```
jontsai: this works, but for style/readability/documentation/prevent human errors, we should do named arguments explicitly instead of implicitly by position:

`autoplay=autoplay`

lib/oembed/utils.py
jontsai: I like how you have it

jontsai: how you currently have it is most readable

lib/oembed/utils.py
ojameso: so you want a line that explicitly states `autoplay = autoplay`?

lib/oembed/utils.py
ojameso: Is this the preferred style when doing strings?

jontsai: it's like riding bike against traffic (wrong) vs with traffic (right) -- driver's aren't trained to dodge bicyclists, but to go around them

in the same way, devs are used to scrolling vertically, not horizontally / having their eyes wrap at lines

ojameso: Will keep in mind.

ojameso:
@jontsai Updated to single quotes/added in explicit setting of autoplay variable.

lib/oembed/utils.py
jontsai: yup, always do single quotes unless some other reason, like needing to use an unescaped single quote.

but even in that situation, in Python I'd use triple quotes `"""`

lib/oembed/utils.py
```diff
@@ -7,12 +7,17 @@
...
+def get_oembed_html(url, autoplay=False):
     """Gets the oEmbed HTML for a URL, if it is an oEmbed type
     """
     oembed_type = get_oembed_type(url)
+
     if oembed_type:
-        html = get_oembed_html_for_service(url, oembed_type)
+        if oembed_type == 'youtube':
+            autoplay = autoplay
+            html = youtube_oembed(url, autoplay)
```
jontsai: no, when you call the function, if the function takes named arguments, invoke the function with name arguments

jontsai: so `html = youtube_oembed(url, autoplay=autoplay)`

when doing assignment in Python, 1 space around the `=`

when funcall, no spaces around the `=`

jontsai: no, do it exactly as I said here: https://github.com/hacktoolkit/django-htk/pull/270#discussion_r202843295

Python knows that LHS = the named arg, and RHS is the variable in scope. no confusion

PR #275, opened 2018-07-24
HTK Toolbar [Emulate User] Extras

ajnassar:
**Trello**: https://trello.com/c/xNtTO7hV


admintools/constants/defaults.py
```diff
@@ -1,3 +1,4 @@
 HTK_COMPANY_EMAIL_DOMAINS = ()
 HTK_COMPANY_OFFICER_EMAILS = ()
 HTK_COMPANY_EMPLOYEE_EMAILS = ()
+HTK_EMULATE_USER_COOKIE_EXPIRATION_MINUTES = 15
```
jontsai: newline above this

templates/htk/fragments/toolbar/base.html
```diff
@@ -11,4 +11,8 @@
 </a>
 {% load htk_tags %}
 {% loadjs '/static/htk/js/toolbar/emulate_user.js' %}
+
+<script type="text/javascript">
```
jontsai: can we rename this file to `toolbar/toolbar.html`

jontsai: also put this above `loadjs` -- the reason this is working is because you have the `$(function() { .... });` wrapper, but otherwise this variable needs to be defined ahead of time

templates/htk/fragments/toolbar/base.html
```diff
@@ -11,4 +11,8 @@
...
+var HTK_EMULATE_USER_COOKIE_EXPIRATION_MINUTES = {% get_htk_setting 'HTK_EMULATE_USER_COOKIE_EXPIRATION_MINUTES' %};
```
jontsai: let's just call it `htk_setting`

templatetags/htk_tags.py
```diff
@@ -206,6 +206,14 @@ def get_django_setting(context, key):
         context[key] = value
     return ''
 
+@register.simple_tag()
+def get_htk_setting(key):
```
jontsai: let's have some namespace fun

just name it `htk_setting`

templatetags/htk_tags.py
```diff
@@ -206,6 +206,14 @@ def get_django_setting(context, key):
...
+    """Retrieves a Django setting and sets it on the context dictionary
+    """
+    from htk.utils import htk_setting
```
jontsai: change the import to

`from htk.utils import htk_setting as _htk_setting`

invoke it as `value = _htk_setting(key)`

admintools/middleware.py
```diff
@@ -23,6 +23,12 @@ def process_request(self, request):
                     emulated_user = get_user_by_id(user_id)
                 elif username:
                     emulated_user = get_user_by_username(username)
-                if emulated_user:
+                if emulated_user and not emulated_user.profile.is_company_officer:
```
jontsai: let's do some refactoring

move `_can_emulate_another_user` to `htk.admintools.utils`

instead of passing in a `request` just pass in the `request.user` and inside the function check `if user.is_authenticated()`

then, add a new function, and replace the `if` condition check here

```
def can_emulate_user(original_user, targeted_user):
    can_emulate = original_user and targeted_user and not targeted_user.profile.is_company_officer
    return can_emulate
```

admintools/middleware.py
```diff
@@ -23,6 +23,12 @@ def process_request(self, request):
...
                     request.original_user = request.user
                     request.user = emulated_user
+
+    def process_response(self, request, response):
+        if self._can_emulate_another_user(request):
```
jontsai: `if not`

jontsai: @ajnassar bump

templatetags/htk_tags.py
```diff
@@ -206,6 +206,12 @@ def get_django_setting(context, key):
         context[key] = value
     return ''
 
+@register.simple_tag()
+def htk_setting(key):
+    from htk.utils import htk_setting
+    value = htk_setting(key)
```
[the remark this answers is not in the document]
ajnassar: Sorry I missed that one, I see it now.

admintools/middleware.py
```diff
@@ -23,6 +13,12 @@ def process_request(self, request):
                     emulated_user = get_user_by_id(user_id)
                 elif username:
                     emulated_user = get_user_by_username(username)
-                if emulated_user:
+                if can_emulate_user(request.user, emulated_user):
                     request.original_user = request.user
                     request.user = emulated_user
+
+    def process_response(self, request, response):
+        if can_emulate_another_user(request.user):
```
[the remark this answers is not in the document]
ajnassar: This logic is fine. If I change it to `if not` it will delete cookies every time we have a successful emulation.

Consider the three possible users.
Bad User (non-existent user): Our request.user will be the original user since we never switched. So if this user happens to be a super user, then we delete the cookie. If not, then this other case doesn't ever happen because it will be the `good user` case
Good User: The request.user will now be our targeted user. This user cannot emulate, so we don't want to delete cookies, let things pass as normal
Employee User: We are able to successfully get this user, but this user is a super user. So we can emulate, therefore we want to remove the cookie.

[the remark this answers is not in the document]
jontsai: @ajnassar you are missing my point, entirely

can you do something like

```
if not hasattr(request, 'original_user') or not can_emulate_another_user(request.user):
```

PR #276, opened 2018-07-27
Feature/htk emulate updates

ajnassar:
https://github.com/jontsai/maskil-www/pull/24
https://github.com/hacktoolkit/htk-css-hacktoolkit/pull/4

### DIFF
--- a/admintools/middleware.py
+++ b/admintools/middleware.py
@@ -34,20 +34,17 @@ class HtkEmulateUserMiddleware(object):
 
                 if targeted_user is None:
                     messages.error(request, 'Cannot Emulate: User does not exist', fail_silently=True)
-                    pass
                 else:
                     if is_allowed_to_emulate(request.user, targeted_user):
                         request.original_user = request.user
                         request.user = targeted_user
                     else:
                         messages.error(request, 'Cannot Emulate: Not allowed to emulate that user', fail_silently=True)
-                        pass
             else:
                 # not attempting to emulate
                 pass
         else:
             messages.error(request, 'Cannot Emulate: Not allowed to emulate that user', fail_silently=True)
-            pass
 
 
     def process_response(self, request, response):
--- a/admintools/middleware.py
+++ b/admintools/middleware.py
@@ -33,20 +33,20 @@ class HtkEmulateUserMiddleware(object):
                     rollbar.report_message('Impossible case: attempting to emulate another user but not specified')
 
                 if targeted_user is None:
-                    messages.error(request, 'Cannot Emulate: User does not exist')
+                    messages.error(request, 'Cannot Emulate: User does not exist', fail_silently=True)
                     pass
                 else:
                     if is_allowed_to_emulate(request.user, targeted_user):
                         request.original_user = request.user
                         request.user = targeted_user
                     else:
-                        messages.error(request, 'Cannot Emulate: Not allowed to emulate that user')
+                        messages.error(request, 'Cannot Emulate: Not allowed to emulate that user', fail_silently=True)
                         pass
             else:
                 # not attempting to emulate
                 pass
         else:
-            messages.error(request, 'Cannot Emulate: Not allowed to emulate that user')
+            messages.error(request, 'Cannot Emulate: Not allowed to emulate that user', fail_silently=True)
             pass
 
 
--- a/admintools/middleware.py
+++ b/admintools/middleware.py
@@ -33,20 +33,20 @@ class HtkEmulateUserMiddleware(object):
                     rollbar.report_message('Impossible case: attempting to emulate another user but not specified')
 
                 if targeted_user is None:
-                    messages.error(request, 'User does not exist')
+                    messages.error(request, 'Cannot Emulate: User does not exist')
                     pass
                 else:
                     if is_allowed_to_emulate(request.user, targeted_user):
                         request.original_user = request.user
                         request.user = targeted_user
                     else:
-                        messages.error(request, 'Not allowed to emulate that user')
+                        messages.error(request, 'Cannot Emulate: Not allowed to emulate that user')
                         pass
             else:
                 # not attempting to emulate
                 pass
         else:
-            messages.error(request, 'Not allowed to emulate that user')
+            messages.error(request, 'Cannot Emulate: Not allowed to emulate that user')
             pass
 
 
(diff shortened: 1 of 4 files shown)


static/htk/js/toolbar/emulate_user.js
```diff
@@ -9,15 +9,23 @@ $(function() {
...
+    }
+
+    function getExpireDate() {
+        var date = new Date();
+        date.setTime(date.getTime() + HTK_EMULATE_USER_COOKIE_EXPIRE_TIME_SECONDS);
+        return date;
+    }
 
     function setPulsatingHtkToolBarHandle() {
         if ($.cookie(EMULATE_USER_ID) || $.cookie(EMULATE_USERNAME)) {
```
jontsai: this could also get shortened using `getActiveCookie()` as well

static/htk/js/toolbar/emulate_user.js
```diff
@@ -27,51 +35,41 @@ $(function() {
...
-    function getCookieName(name) {
-        var cookieNameParts = name.split('-');
-        var cookieName = 'emulate_user_' + cookieNameParts[cookieNameParts.length - 1];
-        return cookieName;
-    }
+    function handleEmulateUserButtonClicked(cookieName) {
+        var cookieVal = emulateUserInput.val();
+        if (cookieVal === '' || (cookieName === EMULATE_USER_ID && isNaN(cookieVal))) {
+            showError();
+            return;
```
jontsai: don't like early `return` statements.

put the rest of the function inside of `else`

static/htk/js/toolbar/emulate_user.js
```diff
@@ -27,51 +35,41 @@ $(function() {
...
-        $.cookie(cookieName, cookieVal, { expires: get_expire_date() });
+        emulateUserMessage.hide();
+        $.cookie(cookieName, cookieVal, { expires: getExpireDate() });
         location.reload();
     }
 
-    function handleEmulateUserStopButtonClicked(cookieName) {
+    function handleEmulateUserStopButtonClicked() {
+        var cookieName;
+        if ($.cookie(EMULATE_USER_ID)) {
```
jontsai: how about we just remove both cookies al the time? simpler, and more bullet proof

don't want some weird condition where somehow two cookies were set

admintools/middleware.py
```diff
@@ -18,7 +18,7 @@ def process_request(self, request):
                     request.user = emulated_user
 
     def process_response(self, request, response):
-        if not (hasattr(request, 'original_user') or can_emulate_another_user(request.user)):
+        if not hasattr(request, 'original_user') or can_emulate_another_user(request.user):
```
jontsai: I think this is wrong.

if you want to do this, you have to express

`not A and not B`

or

`not(A or B)`

ajnassar:
@jontsai I have decided to not include a success message. My two reasons for this is:
1. We are already having a pulsating htk tab
2. We show what user we are emulating in the htk toolbar

Extra note:
I would need to figure out how to only show success message once. If I didn't then the user would always get a success message every time they went to a page which is annoying.

admintools/middleware.py
```diff
@@ -29,31 +33,30 @@ def process_request(self, request):
...
+
     def process_response(self, request, response):
         """Delete user emulation cookies if they should not be set
         """
         original_user = getattr(request, 'original_user', None)
         user = getattr(request, 'user', None)
         is_emulating = original_user is not None and user is not None
-
-        if not(is_emulating or is_allowed_to_emulate_users(user)):
+        if not is_emulating:
```
ajnassar: I changed this back, we need to have a final decision on it, but it was annoying for me to be able to not be able to emulate as a non-existant user or as a super user but appear to be emulating as them in the htk tab and toolbar.

@jontsai I simplified the logic. Since our process_request is doing the grunt of the work, we just need to simply look at the original_user to decide if we want to delete cookies.

static/htk/js/toolbar/emulate_user.js
```diff
@@ -86,20 +81,25 @@ $(function() {
...
         htkToolbarHideButton.click(handleToggleToolbar);
     }
 
     function init() {
-        toggleContainers();
-        toggleForm('user-id');
-        toggleForm('username');
+        toggleForm();
         setPulsatingHtkToolBarHandle();
+        setDjangoAlertMessageToSlide();
```
jontsai: don't like calling it `DjangoAlertMessage`

admintools/middleware.py
jontsai: that seems fine

jontsai: yeah, it makes sense, because you don't want that alert to keep popping up on every page load, I suppose

view_helpers.py
```diff
@@ -15,23 +16,27 @@
...
 
-    # pre-render
-    data['javascripts'] = get_javascripts(template_name, template_prefix=template_prefix)
-    _build_meta_content(data)
-    _build_breadcrumbs(data)
 
-    # render
-    response = render_to_response(template_name, data)
+def render_custom(request, template_name, data=None, template_prefix=''):
+    """Wrapper function for django.shortcuts.render_to_response
```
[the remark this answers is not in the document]
jontsai: also, how is this function used / why is it needed?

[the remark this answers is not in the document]
ajnassar: Okay going to replace `render_to_response` to `render`. For HTK it does the same thing as `render_to_response_custom` adding the same stuff to the context. But it is different in that it is django's `render` instead of `render_to_response`.

static/htk/js/toolbar/emulate_user.js
```diff
@@ -86,20 +81,25 @@ $(function() {
         }
     }
 
+    function setDjangoAlertMessageToSlide() {
+        // If Django messaging is enabled, this will make the message slide up after 5 seconds
+        $(".alert.alert-dismissible").fadeTo(5000, 500).slideUp(500, function(){
+            $(".alert.alert-dismissible").slideUp(500);
```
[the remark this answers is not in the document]
ajnassar: It hides the element.

view_helpers.py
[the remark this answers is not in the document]
jontsai: @ajnassar 

it looks like `render` is the newer/better replacement for `render_to_response`

Per this [documentation](https://docs.djangoproject.com/en/1.11/topics/http/shortcuts/#render-to-response), its only difference is in also accepting the `request` as an argument, but otherwise all the same.

Can you put this function above? We should gradually start replacing everything with `render`

PR #277, opened 2018-07-31
Remove message when user is not allowed to emulate

ajnassar:
@jontsai We are messaging all the time for user's that aren't allowed to emulate. They should just simply pass through the middleware.

### DIFF
--- a/admintools/middleware.py
+++ b/admintools/middleware.py
@@ -44,7 +44,7 @@ class HtkEmulateUserMiddleware(object):
                 # not attempting to emulate
                 pass
         else:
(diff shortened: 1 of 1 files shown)


admintools/middleware.py
```diff
@@ -44,7 +44,8 @@ def process_request(self, request):
                 # not attempting to emulate
                 pass
         else:
-            messages.error(request, 'Cannot Emulate: Not allowed to emulate that user', fail_silently=True)
+            # is not allowed to emulate users
```
jontsai: comment should be `is not allowed or is not attempting to emulate users`

PR #281, opened 2018-08-22
Add has_permission simple_tag

ajnassar:
## Status
**READY**

### DIFF
--- a/templatetags/htk_tags.py
+++ b/templatetags/htk_tags.py
@@ -219,6 +219,26 @@ def get_request_duration():
     duration = timer.duration()
     return duration
 
+@register.simple_tag(takes_context=True)
+def has_permission(context, codename):
+    user = context.get('user', None)
+    # UserModel = get_user_model()
+    # user = UserModel.objects.get(id=user_id)
+    has_permission = False
+
+    # Check if user has permission explicitly given
+    for permission in user.user_permissions.all():
+        if permission.codename == codename:
+            has_permission = True
+
+    # Check if user has permission from group
+    for group in user.groups.all():
+        for permission in group.permissions.all():
+            if permission.codename == codename:
+                has_permission = True
+
+    return has_permission
+
 ##
 # Load Assets
 
--- a/templatetags/htk_tags.py
+++ b/templatetags/htk_tags.py
@@ -3,6 +3,7 @@ import re
 import urllib
 
 from django import template
+from django.contrib.auth.models import Permission
 from django.template.defaultfilters import stringfilter
 from django.urls import reverse
 from django.utils.safestring import mark_safe
@@ -221,21 +222,13 @@ def get_request_duration():
 
 @register.simple_tag(takes_context=True)
 def has_permission(context, codename):
-    user = context.get('user', None)
-    # UserModel = get_user_model()
-    # user = UserModel.objects.get(id=user_id)
-    has_permission = False
-
-    # Check if user has permission explicitly given
-    for permission in user.user_permissions.all():
-        if permission.codename == codename:
-            has_permission = True
-
-    # Check if user has permission from group
-    for group in user.groups.all():
-        for permission in group.permissions.all():
-            if permission.codename == codename:
-                has_permission = True
+    request_meta_data = context.get('request', None)
+    request = request_meta_data.get('request', None)
+    user = request.user
+    permission = Permission.objects.get(codename=codename)
+    app_label = permission.content_type.app_label
+    permission_key = '%s.%s' % (app_label, codename)
+    has_permission = user.has_perm(permission_key)
 
     return has_permission
 
--- a/templatetags/htk_tags.py
+++ b/templatetags/htk_tags.py
@@ -274,17 +274,11 @@ def is_editable_by_context_user(context, obj):
     return is_editable
 
 @register.simple_tag(takes_context=True)
-def has_permission(context, permission_codename):
+def has_permission(context, permission_key):
     request = context.get('request', {}).get('request', None)
     user = request.user
     if request and user.is_authenticated():
-        try:
-            permission = Permission.objects.get(codename=permission_codename)
-            app_label = permission.content_type.app_label
-            permission_key = '%s.%s' % (app_label, permission_codename)
-            has_permission = user.has_perm(permission_key)
-        except Permission.DoesNotExist:
-            has_permission = False
+        has_permission = user.has_perm(permission_key)
     else:
         has_permission = False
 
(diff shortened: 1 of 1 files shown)


templatetags/htk_tags.py
```diff
@@ -219,6 +219,26 @@ def get_request_duration():
     duration = timer.duration()
     return duration
 
+@register.simple_tag(takes_context=True)
+def has_permission(context, codename):
+    user = context.get('user', None)
+    # UserModel = get_user_model()
```
jontsai: remove

templatetags/htk_tags.py
```diff
@@ -219,6 +219,26 @@ def get_request_duration():
...
+    # user = UserModel.objects.get(id=user_id)
```
jontsai: from the docs you sent me:

https://docs.djangoproject.com/en/1.11/topics/auth/default/#default-permissions

You just need to do `user.has_perm(codename)`, I believe

templatetags/htk_tags.py
```diff
@@ -219,6 +220,18 @@ def get_request_duration():
     duration = timer.duration()
     return duration
 
+@register.simple_tag(takes_context=True)
```
jontsai: move this to the section with `ACL Tags`

templatetags/htk_tags.py
```diff
@@ -219,6 +220,18 @@ def get_request_duration():
...
+def has_permission(context, codename):
```
jontsai: rename variable to `permission_codename`

templatetags/htk_tags.py
```diff
@@ -219,6 +220,18 @@ def get_request_duration():
...
+    request_meta_data = context.get('request', None)
+    request = request_meta_data.get('request', None)
```
jontsai: replace two lines with defensive coding style

`request = context.get('request', {}).get('request', None)`

templatetags/htk_tags.py
```diff
@@ -219,6 +220,18 @@ def get_request_duration():
...
+    user = request.user
```
jontsai: add a check for `if request` and/or `if user` (defensive coding)

if not, `has_permission = False`

templatetags/htk_tags.py
```diff
@@ -219,6 +220,18 @@ def get_request_duration():
...
+    permission = Permission.objects.get(codename=codename)
+    app_label = permission.content_type.app_label
+    permission_key = '%s.%s' % (app_label, codename)
+    has_permission = user.has_perm(permission_key)
```
jontsai: this feels... circuitous

do they support multiple modes/syntax? like does it just take the permission object, since you already have it? seems pedantic to have to build the key again when you already have the object (and they obviously would have to look up the object again)

ajnassar: I am not sure what `obj` refers to. The model obj? the permission obj? In any case the docs say you need to pass in `perm` which is a string such that `app_label.perm_code`

templatetags/htk_tags.py
```diff
@@ -285,6 +273,20 @@ def is_editable_by_context_user(context, obj):
         is_editable = False
     return is_editable
 
+@register.simple_tag(takes_context=True)
+def has_permission(context, permission_codename):
+    request = context.get('request', {}).get('request', None)
+    if request and request.user:
```
jontsai: `user` will always be there on a Django request, so we don't need to check for this

It's whether the user is `user.is_authenticated` or not

templatetags/htk_tags.py
```diff
@@ -285,6 +273,20 @@ def is_editable_by_context_user(context, obj):
...
+        user = request.user
+        permission = Permission.objects.get(codename=permission_codename)
```
jontsai: what if you pass in a garbage codename, and it doesn't exist?

seems circuitous and wasteful to get the permission object, only to extract the `app_label`

would like a better way

templatetags/htk_tags.py
ajnassar: I see, I made the assumption that this is behind admintools door, but this tag can be used anywhere. So we only want to do this when user is authenticated.

jontsai: Django `request.user` is always an object. it can be an `AnonymousUser`

jontsai: https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#anonymoususer-object

templatetags/htk_tags.py
jontsai: is it possible to pass in multiple parameters? in this case, also pass in the `app_label`

or try to get `app_label` from somewhere else, if it's available

templatetags/htk_tags.py
```diff
@@ -272,6 +273,20 @@ def is_editable_by_context_user(context, obj):
         is_editable = False
     return is_editable
 
+@register.simple_tag(takes_context=True)
+def has_permission(context, permission_codename):
```
jontsai: what if you just pass in the full `permission_key` instead of `permission_codename`? saves the additional lookup and stringbuilding

templatetags/htk_tags.py
ajnassar: I get this when I search for a Permission with non-existent codename
image

I will wrap this in a try/except

ajnassar: No, the docs say to pass in a string `app_label.permission_codename`.

templatetags/htk_tags.py
ajnassar: This would mean the user needs to know the app_label as well. I think it will be easier if the user didn't have to worry about that.

ajnassar: I made up the term, I tried looking for something. Would you be okay if we had two params in this case to be extra clear. `app_label` and `permission_codename` or should we just go with `permission_key`?

jontsai: the latter. one.

Django calls it `perm`, which is rather inane

jontsai: but feel free to call it `perm` as well, if that makes you feel better

templatetags/htk_tags.py
```diff
@@ -3,6 +3,7 @@
 import urllib
```
jontsai: @ajnassar can you check whether `static/htk/less` was correctly modified or not?

templatetags/htk_tags.py
```diff
@@ -3,6 +3,7 @@
...
 
 from django import template
+from django.contrib.auth.models import Permission
```
jontsai: remove unused import

PR #296, opened 2019-07-15
Adds get_distinct_members to Organization

brywong:


apps/organizations/models.py
```diff
@@ -15,7 +15,7 @@
 from htk.models import AbstractAttributeHolderClassFactory
 from htk.models import HtkBaseModel
 from htk.utils import htk_setting
-
```
jontsai: could you alphabetize imports?

PR #306, opened 2021-03-03
Added user timezone form

sakthivel-axim:
Added timezone form to store timezone in the user profile.

Reference PR https://github.com/talentral/www/pull/1273

# Trello Ticket(s)
https://trello.com/c/a91sS3IA

### DIFF
--- a/apps/accounts/forms/update.py
+++ b/apps/accounts/forms/update.py
@@ -156,7 +156,7 @@ class TimeZoneForm(AbstractModelInstanceUpdateForm):
         user = super(TimeZoneForm, self).save(request, *args, **kwargs)
         user_profile = user.profile
         timezone = request.POST.get('timezone', '')
-        default_timezone = htk_setting('HTK_DEFAULT_TIMEZONE')
+        default_timezone = htk_setting('HTK_DEFAULT_TIMEZONE', 'UTC')
         user_profile.timezone = timezone or default_timezone
         user_profile.save()
         return user
--- a/apps/accounts/forms/update.py
+++ b/apps/accounts/forms/update.py
@@ -13,10 +13,14 @@ from django.contrib.auth.forms import SetPasswordForm
 # HTK Imports
 from htk.apps.accounts.emails import password_changed_email
 from htk.forms import AbstractModelInstanceUpdateForm
-from htk.forms.utils import set_input_attrs
-from htk.forms.utils import set_input_placeholder_labels
-from htk.utils import htk_setting
-from htk.utils import resolve_model_dynamically
+from htk.forms.utils import (
+    set_input_attrs,
+    set_input_placeholder_labels,
+)
+from htk.utils import (
+    htk_setting,
+    resolve_model_dynamically,
+)
 from htk.utils.geo import get_us_state_abbreviation_choices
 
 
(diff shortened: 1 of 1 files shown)


apps/accounts/forms/update.py
```diff
@@ -129,3 +133,29 @@ def save(self, user, *args, **kwargs):
...
+        super(TimeZoneForm, self).__init__(user, *args, **kwargs)
+
+        self.fields['timezone'].initial = user_profile.timezone
+        self.fields['timezone'].label = ''
+
+    def save(self, request, *args, **kwargs):
+        user = super(TimeZoneForm, self).save(request, *args, **kwargs)
+        user_profile = user.profile
+        timezone = request.POST.get('timezone', '')
+        user_profile.timezone = timezone if timezone else 'America/Los_Angeles'
```
jontsai, line 159: The fallback here should be something like:

```
from htk.utils import htk_setting
...
default_timezone = htk_setting('HTK_DEFAULT_TIMEZONE')
user_profile.timezone = timezone or default_timezone
```

jontsai (changes requested):

apps/accounts/forms/update.py
sakthivel-axim: Done

apps/accounts/forms/update.py
```diff
@@ -129,3 +133,30 @@ def save(self, user, *args, **kwargs):
...
+
+        self.fields['timezone'].initial = user_profile.timezone
+        self.fields['timezone'].label = ''
+
+    def save(self, request, *args, **kwargs):
+        user = super(TimeZoneForm, self).save(request, *args, **kwargs)
+        user_profile = user.profile
+        timezone = request.POST.get('timezone', '')
+        default_timezone = htk_setting('HTK_DEFAULT_TIMEZONE', 'UTC')
+        user_profile.timezone = timezone or default_timezone
```
jontsai, line 160:
```
default_timezone = ...
timezone = request.POST.get('timezone', default_timezone)
user_profile.timezone = timezone
```

jontsai (changes requested):

apps/accounts/forms/update.py
```diff
@@ -129,3 +133,30 @@ def save(self, user, *args, **kwargs):
...
+
+    def __init__(self, instance, *args, **kwargs):
+        """Initialization for TimeZoneForm.
+        """
+        user = instance
+        user_profile = user.profile
+        super(TimeZoneForm, self).__init__(user, *args, **kwargs)
+
+        self.fields['timezone'].initial = user_profile.timezone
+        self.fields['timezone'].label = ''
```
jontsai, line 153: I think these two lines should probably be removed

apps/accounts/forms/update.py
```diff
@@ -129,3 +133,30 @@ def save(self, user, *args, **kwargs):
...
+        fields = (
+            'timezone',
+        )
+
+    def __init__(self, instance, *args, **kwargs):
+        """Initialization for TimeZoneForm.
+        """
+        user = instance
+        user_profile = user.profile
+        super(TimeZoneForm, self).__init__(user, *args, **kwargs)
```
jontsai, line 150: `super(TimeZoneForm, self).__init__(user_profile, *args, **kwargs)`

apps/accounts/forms/update.py
```diff
@@ -129,3 +133,30 @@ def save(self, user, *args, **kwargs):
...
+        model = UserProfile
+        fields = (
+            'timezone',
+        )
+
+    def __init__(self, instance, *args, **kwargs):
+        """Initialization for TimeZoneForm.
+        """
+        user = instance
+        user_profile = user.profile
```
jontsai, line 149: Just pass in `user_profile` in the constructor

```
user_profile = instance
```

jontsai (changes requested):

apps/accounts/forms/update.py
sakthivel-axim: The empty(-----) also one of the options in the timezone field of the UserProfile, 
image

which is due to blank attribute is set to true in the following line, 
https://github.com/hacktoolkit/django-htk/blob/4441c12d1b5e8592a4039973754d33969e996cc3/apps/accounts/models.py#L69

If we keep the above code (https://github.com/hacktoolkit/django-htk/blob/44ea9f6c244bcfc5b50aee858c9217e43085d265/apps/accounts/forms/update.py#L158-L160) then it will set it to default timezone when the timezone is empty otherwise, We need to change to the field to required by setting blank to False here  https://github.com/hacktoolkit/django-htk/blob/4441c12d1b5e8592a4039973754d33969e996cc3/apps/accounts/models.py#L69


Shall I keep this code?

apps/accounts/forms/update.py
sakthivel-axim: Done

apps/accounts/forms/update.py
sakthivel-axim: Done

jontsai (approved):

PR #308, opened 2021-04-19
Add levenshtein distance for autocorrection

sakthivel-axim:
**READY**

## Description
Added functions for word autocorrection using `Levenshtein Distance` algorithm.

## Trello Ticket(s)
https://trello.com/c/rXew3TGN


utils/text/levenshtein.py
```diff
@@ -0,0 +1,61 @@
...
+    for title in job_titles:
+        job_title_distance = getLevenshteinDistance(job_title, title.strip())
+        if job_title_distance >= 10:
+            job_title_distance = 9
+        job_titles_list.append(str(int(job_title_distance)) + "-" + title.strip())
+    job_titles_list.sort()
+
+    for i in range(MAX_MATCHED_TITLES):
+        matched_job_titles_list.append(job_titles_list[i].split("-")[1])
+    return matched_job_titles_list[0]
```
jontsai, line 30: This does not belong in this file

jontsai (changes requested):

utils/text/levenshtein.py
sakthivel-axim: Done.

utils/text/levenshtein.py
```diff
@@ -0,0 +1,37 @@
+# Third Party (PyPI) Imports
+import numpy
+
+
+def getLevenshteinDistance(job_title, title_from_json):
```
jontsai, line 5: rename signature to be more generic:

`def levenshtein(w1, w2)`

utils/text/levenshtein.py
```diff
@@ -0,0 +1,37 @@
...
+    """The Levenshtein distance algorithm that compares two words.
+
+    Returns a numeric value representing the distance between two words.
+    """
+    a = 0
+    b = 0
+    c = 0
+    title_distance = numpy.zeros((len(job_title) + 1, len(title_from_json) + 1))
```
jontsai, line 13: `edit_distance`

jontsai (changes requested):

utils/text/levenshtein.py
```diff
@@ -0,0 +1,37 @@
+# Third Party (PyPI) Imports
+import numpy
+
+
+def getLevenshteinDistance(job_title, title_from_json):
+    """The Levenshtein distance algorithm that compares two words.
+
+    Returns a numeric value representing the distance between two words.
+    """
```
jontsai, line 9:
```
"""The Levenshtein distance algorithm that compares two words

https://en.wikipedia.org/wiki/Levenshtein_distance

Returns an `int` representing the edit distance between two words
"""
```

jontsai (changes requested):

utils/text/levenshtein.py
sakthivel-axim: Done

utils/text/levenshtein.py
sakthivel-axim: Done

utils/text/levenshtein.py
sakthivel-axim: Done

utils/text/levenshtein.py
```diff
@@ -0,0 +1,39 @@
...
+
+    for x in range(len(w1) + 1):
+        edit_distance[x][0] = x
+
+    for y in range(len(w2) + 1):
+        edit_distance[0][y] = y
+
+    for x in range(1, len(w1) + 1):
+        for y in range(1, len(w2) + 1):
+            if (w1[x-1] == w2[y-1]):
```
jontsai, line 25: spaces around `x - 1` and `y - 1`

utils/text/levenshtein.py
```diff
@@ -0,0 +1,39 @@
...
+                c = edit_distance[x - 1][y - 1]
+
+                if (a <= b and a <= c):
+                    edit_distance[x][y] = a + 1
+                elif (b <= a and b <= c):
+                    edit_distance[x][y] = b + 1
+                else:
+                    edit_distance[x][y] = c + 1
+
+    return edit_distance[len(w1)][len(w2)]
```
jontsai, line 39: assign a temp variable before returning

```
result = edit_distance[len(w1)][len(w2)]
return result
```

utils/text/levenshtein.py
```diff
@@ -0,0 +1,39 @@
+# Third Party (PyPI) Imports
```
jontsai, line 1: 1. Can we rename this file to `algorithms.py` and the function name to `levenshtein_distance`
2. Can we also add `numpy` to `requirements.txt`: https://github.com/hacktoolkit/django-htk/blob/master/requirements.txt; Leave it unpinned
3. Inside of https://github.com/hacktoolkit/django-htk/blob/master/utils/text/__init__.py, can you add:
    ```
    from htk.utils.text.algorithms import levenshtein_distance
    ```
    So that it can be imported as `from htk.utils.text import levenshtein_distance`
4. I think this implementation is potential suboptimal, in that it constructs an `m * n` matrix. Can we use this implementation instead: https://en.wikipedia.org/wiki/Levenshtein_distance#Iterative_with_two_matrix_rows (also update the Wikipedia URL in the docstring)
5. Would prefer to use `deletion_cost`, `insertion_cost`, `substitution_cost` as intermediate variable names, instead of `a`, `b`, `c`

jontsai (changes requested):

utils/text/levenshtein.py
sakthivel-axim: 1. Renamed file and function
2. Added numpy to requirements.txt
3. Added import to utils/text/__init__.py
4. The results of the algorithm from the link https://en.wikipedia.org/wiki/Levenshtein_distance#Iterative_with_two_matrix_rows is not valid and keeping existing implementation as it is. Will check other optimal algorithms to get the valid distance.
5. Renamed the variables

utils/text/levenshtein.py
sakthivel-axim: Done

utils/text/levenshtein.py
sakthivel-axim: Done

aarthi-axim (approved):

utils/text/algorithms.py
```diff
@@ -0,0 +1,36 @@
...
+        for y in range(1, len(w2) + 1):
+            if (w1[x - 1] == w2[y - 1]):
+                edit_distance[x][y] = edit_distance[x - 1][y - 1]
+            else:
+                insertion_cost = edit_distance[x][y - 1]
+                deletion_cost = edit_distance[x - 1][y]
+                substitution_cost = edit_distance[x - 1][y - 1]
+
+                min_cost = min(deletion_cost, insertion_cost, substitution_cost)
+                edit_distance[x][y] = min_cost + 1
```
jontsai, line 33:
```suggestion
                insertion_cost = edit_distance[x][y - 1] + 1
                deletion_cost = edit_distance[x - 1][y] + 1
                substitution_cost = edit_distance[x - 1][y - 1] + 1

                edit_distance[x][y] = min(deletion_cost, insertion_cost, substitution_cost)
```

utils/text/algorithms.py
```diff
@@ -0,0 +1,36 @@
+# Third Party (PyPI) Imports
+import numpy
+
+
+def levenshtein_distance(w1, w2):
+    """The Levenshtein distance algorithm that compares two words
+
```
jontsai, line 7: Let's also include the Wikipedia link:

`https://en.wikipedia.org/wiki/Levenshtein_distance`

jontsai (changes requested):

jontsai (approved):

PR #313, opened 2021-05-14
Fixed to get email only confirmed email object exists

sakthivel-axim:
## Status
**READY**

## Description
Fixed to get email only confirmed email object exists.

https://trello.com/c/x61pUORN

### DIFF
--- a/apps/accounts/models.py
+++ b/apps/accounts/models.py
@@ -238,8 +238,8 @@ class BaseAbstractUserProfile(HtkBaseModel, UserAttributeHolder, HtkCompanyUserM
         else:
             confirmed_emails = self.get_confirmed_emails()
             if confirmed_emails.exists():
-
-                email = confirmed_emails.first().email
+                confirmed_emails = confirmed_emails.first()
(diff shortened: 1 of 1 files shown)


aarthi-axim (approved):

apps/accounts/models.py
```diff
@@ -238,8 +238,8 @@ def confirmed_email(self):
         else:
             confirmed_emails = self.get_confirmed_emails()
             if confirmed_emails.exists():
-
-                email = confirmed_emails.first().email
+                confirmed_emails = confirmed_emails.first()
```
jontsai, line 241: This shouldn't be necessary, because this is already inside of `if confirmed_emails.exists()`

jontsai (changes requested):

apps/accounts/models.py
sakthivel-axim: Ok. But not sure how `confirmed_emails.exists()` succeeded and failed to get `confirmed_emails.first().email`. We can close this PR if not needed.

sakthivel-axim:
Closing this as per review comment.

PR #315, opened 2021-05-20
Add error handle for webhook topics

aarthi-axim:
## Status
**READY**

## Description
Added error handle to webhook topics.

### DIFF
--- a/lib/dynamic_screening_solutions/api.py
+++ b/lib/dynamic_screening_solutions/api.py
@@ -458,6 +458,9 @@ class Htk321FormsAPI(object):
         except Exception:
             webhook_topics = []
 
+        if 'error' in webhook_topics and webhook_topics['error'] is True:
+            webhook_topics = []
+
         return webhook_topics
 
     def create_webhook(self, company_id, url, topic_ids):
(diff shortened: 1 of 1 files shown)


lib/dynamic_screening_solutions/api.py
```diff
@@ -458,6 +458,9 @@ def get_webhook_topics(self, company_id):
         except Exception:
             webhook_topics = []
 
+        if 'error' in webhook_topics and webhook_topics['error'] is True:
+            webhook_topics = []
+
```
jontsai, line 463: Let's move these 3 lines into the `try` block

jontsai (changes requested):

lib/dynamic_screening_solutions/api.py
```diff
@@ -458,6 +458,9 @@ def get_webhook_topics(self, company_id):
         except Exception:
             webhook_topics = []
 
+        if 'error' in webhook_topics and webhook_topics['error'] is True:
+            webhook_topics = []
```
jontsai, line 462:
```suggestion
            if webhook_topics.get('error'):
                webhook_topics = []
        except Exception:
            webhook_topics = []

```

We can move it into the `try` block and simply the expression to just do a `.get`

jontsai (changes requested):

lib/dynamic_screening_solutions/api.py
aarthi-axim: Done

jontsai (approved):
LGTM

PR #319, opened 2021-11-04
features/django-htk/encoding-unicode-ascii-characters

YosefAshenafi:
Encoding Unicode ascii characters

This fix enables Unicode ascii's encoding with str() which were brought an error "UnicodeEncodeError: 'ascii' codec can't encode character u'\xa0' in position 20: ordinal not in range(128)" to be encoded with 'utf-8'.

### DIFF
--- a/cache/classes.py
+++ b/cache/classes.py
@@ -186,7 +186,7 @@ class CustomCacheScheme(object):
 
         Can be overridden by the subclass
         """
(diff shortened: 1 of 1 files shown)


jontsai:
#328 is a better approach, but in the end, deciding to keep the code as is so as to not encounter unpredictable behavior. Pass in a regular ASCII string as `prekey`

PR #321, opened 2021-12-07
Add Neighborhood field in BasePostalAddress

aarthi-axim:
Added `Neighborhood` field in `BasePostalAddress`

### DIFF
--- a/apps/addresses/models.py
+++ b/apps/addresses/models.py
@@ -16,19 +16,22 @@ class BasePostalAddress(AbstractGeolocation):
     This object is always referenced by a foreign key from another object
 
     Storing address as a separate model is a best practice
-    Other than that, there are none. Parsing postal addresses and (even human names, for that matter) are notoriously difficult (not even taking international addresses into consideration)
+    Other than that, there are none. Parsing postal addresses and (even human names, for that matter) are
+    notoriously difficult (not even taking international addresses into consideration)
 
     http://stackoverflow.com/questions/310540/best-practices-for-storing-postal-addresses-in-a-database-rdbms
 
-    The safest way would be to have a free-form field to enter whatever address, and then an algorithm that can be maintained over time that parses the address into structured fields.
+    The safest way would be to have a free-form field to enter whatever address, and then an algorithm that can
+    be maintained over time that parses the address into structured fields.
 
     Right now, go with the "good enough" approach
     """
     name = models.CharField(max_length=64, blank=True)
     street = models.CharField(max_length=128, blank=True)
+    neighborhood = models.CharField(max_length=64, blank=True)
     city = models.CharField(max_length=64, blank=True)
     state = models.CharField(max_length=2, blank=True)
-    zipcode = models.CharField(max_length=5, blank=True)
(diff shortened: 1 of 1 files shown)


apps/addresses/models.py
```diff
@@ -16,19 +16,22 @@ class BasePostalAddress(AbstractGeolocation):
     This object is always referenced by a foreign key from another object
 
     Storing address as a separate model is a best practice
-    Other than that, there are none. Parsing postal addresses and (even human names, for that matter) are notoriously difficult (not even taking international addresses into consideration)
```
aarthi-axim, line 19: ## REVIEWER NOTE:
Lint fix

apps/addresses/models.py
```diff
@@ -16,19 +16,22 @@ class BasePostalAddress(AbstractGeolocation):
...
 
     Right now, go with the "good enough" approach
     """
     name = models.CharField(max_length=64, blank=True)
     street = models.CharField(max_length=128, blank=True)
+    neighborhood = models.CharField(max_length=64, blank=True)
     city = models.CharField(max_length=64, blank=True)
     state = models.CharField(max_length=2, blank=True)
-    zipcode = models.CharField(max_length=5, blank=True)
+    zipcode = models.CharField(max_length=16, blank=True)
```
aarthi-axim, line 34: ## REVIEWER NOTE:
Changed `max_length` to make it compatible with `zippy model`

apps/addresses/models.py
```diff
@@ -46,7 +49,7 @@ def __str__(self):
         return value
 
     def clone(self):
-        address_clone = PostalAddress.objects.create(
```
aarthi-axim, line 49: ## REVIEWER NOTE:
Cloning should be done for self model(`BasePostalAddress`)

apps/addresses/models.py
```diff
@@ -120,7 +123,7 @@ def get_static_google_map_image_url(self):
         height = 234
         str_value = self.__str__()
         base_url = 'http://maps.googleapis.com/maps/api/staticmap?'
-        query = urllib.urlencode(
```
aarthi-axim, line 123: ## REVIEWER NOTE:
`urllib` from `six.moves` has no `urlencode` instead it should be imported from `urllib.parse`

apps/addresses/models.py
jontsai: Somewhat leaning toward keeping it as is for now, and modifying it in the subclass

apps/addresses/models.py
```diff
@@ -46,7 +49,7 @@ def __str__(self):
         return value
 
     def clone(self):
-        address_clone = PostalAddress.objects.create(
+        address_clone = BasePostalAddress.objects.create(
```
jontsai, line 52: This is not correct. I think It would need to be changed to something like `self.__class__.objects.create(...)`

jontsai (changes requested):

apps/addresses/models.py
aarthi-axim: Done

jontsai (approved):

PR #323, opened 2022-01-18
Add stripe charge model

sakthivel-axim:
Add stripe charge model

Reference: https://stripe.com/docs/api/charges

### DIFF
--- a/lib/stripe_lib/models.py
+++ b/lib/stripe_lib/models.py
@@ -484,7 +484,7 @@ class AbstractStripeCustomerHolder(models.Model):
 
 class BaseStripeCharge(models.Model):
     stripe_id = models.CharField(max_length=255)
-    amount = models.DecimalField(max_digits=9, decimal_places=2, default=0.0)
+    amount = models.PositiveIntegerField(default=0)
 
     class Meta:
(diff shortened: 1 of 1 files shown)


lib/stripe_lib/models.py
```diff
@@ -480,3 +480,15 @@ def add_or_replace_credit_card(self, card):
             customer = self.create_stripe_customer(card=card)
             was_added_or_replaced = customer is not None
         return was_added_or_replaced
+
+
+class BaseStripeCharge(models.Model):
+    stripe_id = models.CharField(max_length=255)
+    amount = models.PositiveIntegerField(default=0)
+
```
jontsai, line 488: This is missing at least the `currency` field

jontsai (changes requested):
I fundamentally don't agree with storing `StripeCharge` in the database, because there are too many potential charges. I strongly prefer a live API call to be made to the Stripe API, and let that be used as the source of truth for charges.

If absolutely needed, I recommend  storing just the `stripe_charge_id` for individual purchases, etc. (Also, if needed, store it along with the `stripe_customer_id`.)

PR #325, opened 2022-02-08
Adds `json_encode` to `BasePostalAddress` of address app

goztrk:


apps/addresses/models.py
```diff
@@ -65,6 +65,25 @@ def clone(self):
         )
         return address_clone
 
+    def json_encode(self):
+        json_encode = {
```
jontsai, line 69: please rename the variable to something like `payload` or `data`

jontsai (changes requested):

apps/addresses/models.py
goztrk: Done

jontsai (approved):

PR #327, opened 2022-02-20
Prelaunch signup form change

YosefAshenafi:
Prelaunch Signup for change. 
- Signup button to Subscribe
- Show labels all time
- Input height padding increased


jontsai:
The template can and should be overridden locally.

PR #342, opened 2022-09-01
Add support to unicode type for Subscription data

aarthi-axim:
Provide support for `unicode` type subscription data

### DIFF
--- a/lib/stripe_lib/models.py
+++ b/lib/stripe_lib/models.py
@@ -358,9 +358,11 @@ class BaseStripeCustomer(BaseStripeModel):
             item = {
                 'id': stripe_subscription['items']['data'][0]['id'],
             }
-            if type(new_price_or_plan) == str:
+            if isinstance(new_price_or_plan, str):
                 item['price'] = new_price_or_plan
-            elif type(new_price_or_plan) == dict:
+            elif isinstance(new_price_or_plan, type(u'')):
+                item['price'] = new_price_or_plan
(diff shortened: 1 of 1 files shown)


lib/stripe_lib/models.py
```diff
@@ -358,9 +358,11 @@ def change_subscription_plan(self, subscription_id, new_price_or_plan, prorate=T
             item = {
                 'id': stripe_subscription['items']['data'][0]['id'],
             }
-            if type(new_price_or_plan) == str:
+            if isinstance(new_price_or_plan, str):
                 item['price'] = new_price_or_plan
-            elif type(new_price_or_plan) == dict:
+            elif isinstance(new_price_or_plan, type(u'')):
```
aarthi-axim, line 363: ## Reviewer Note:
Supports `unicode` type as compatible for python 2 and python 3

Used `isinstance` instead of comparing `type` on both RHS and LHS to avoid lint issue

lib/stripe_lib/models.py
```diff
@@ -358,9 +358,11 @@ def change_subscription_plan(self, subscription_id, new_price_or_plan, prorate=T
...
+                item['price'] = new_price_or_plan
+            elif isinstance(new_price_or_plan, dict):
```
aarthi-axim, line 365: ## Reviewer Note:
Used `isinstance` instead of `type` inorder to follow same pattern for the conditional checks

jontsai (approved):

PR #348, opened 2022-12-09
Fixed issues with 321Forms API response

aarthi-axim:
## STATUS ##
READY

## DESCRIPTION ##
Fixed issues with 321Forms API responses

### DIFF
--- a/lib/dynamic_screening_solutions/api.py
+++ b/lib/dynamic_screening_solutions/api.py
@@ -108,6 +108,20 @@ class Htk321FormsAPI(object):
             extra_data=extra_data
         )
 
+    def handle_response(self, response):
+        try:
+            response_json = response.json()
+            response_json = None if type(response_json) == dict and response_json.get('error') else response_json
+        except ValueError:
+            response_json = None
+
+        if response_json is None:
+            self.handle_bad_response(response)
+        else:
+            pass
+
+        return response_json
+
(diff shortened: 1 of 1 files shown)


lib/dynamic_screening_solutions/api.py
```diff
@@ -484,6 +484,7 @@ def get_webhooks_by_company(self, company_id):
         request_url = self.get_request_url(resource_path=resource_path)
 
         response_json = self.request_get(request_url)
+        response_json = response_json or []
```
jontsai, line 487:
```suggestion
        response_json = self.request_get(request_url) or []
```

lib/dynamic_screening_solutions/api.py
```diff
@@ -108,6 +108,20 @@ def handle_bad_response(self, response):
...
 
+    def handle_response(self, response):
+        try:
+            response_json = response.json()
+            response_json = None if type(response_json) == dict and response_json.get('error') else response_json
+        except ValueError:
+            response_json = None
+
+        if response_json is None:
+            self.handle_bad_response(response)
```
jontsai, line 119: Needs a `handle_error_response` method for handling an actual 200 response with `{‘error’ : … }` in the response as valid JSON, rather than 1 single `handle_bad_response` (non-valid JSON returned).

jontsai (changes requested):

jontsai (approved):

PR #363, opened 2023-02-28
💄🩹 fix: styling response problems on feature flag page

goztrk:
Fixed responsive styling problems on Feature Flag page.

- If screen size is `>= 1400px` it will show `3` columns
- If screen size is `>= 1200px` it will show `2` columns
- If screen size is  `< 1200px` it will show `1` column


jontsai (approved):
LGTM

PR #364, opened 2023-03-19
Add guard to prelaunch get_or_create_by_email method

goztrk:
`PrelaunchSignup.get_or_create_by_email` method now works correctly even if there are multiple records with same email address.

If there are same records which early access enabled for them, it picks it over other ones.

### DIFF
--- a/apps/prelaunch/models.py
+++ b/apps/prelaunch/models.py
@@ -10,10 +10,7 @@ from django.db import models
 
 # HTK Imports
 from htk.apps.prelaunch.emails import prelaunch_email
-from htk.utils import (
-    htk_setting,
-    utcnow,
-)
+from htk.utils import htk_setting
 from htk.utils.notifications import notify
 from htk.utils.request import get_current_request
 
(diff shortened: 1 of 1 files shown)


apps/prelaunch/models.py
```diff
@@ -66,7 +63,16 @@ def get_or_create_by_email(
...
-            prelaunch_signup = cls.objects.get(email=email, site=site)
+            prelaunch_signups = cls.objects.filter(email=email, site=site)
+            prelaunch_signups_with_early_access = prelaunch_signups.filter(
+                early_access=True
+            )
+            prelaunch_signup = (
+                prelaunch_signups_with_early_access.first()
+                if prelaunch_signups_with_early_access.count() > 0
+                else prelaunch_signups.first()
+            )
```
jontsai, line 74:
```suggestion
            prelaunch_signup = (
                prelaunch_signups_with_early_access.first()
                or prelaunch_signups.first()
            )
```

- `.first()` will return `None` if it does not exist, so it is always safe to call
- `.exists()` is a better pattern and preferred over `.count() > 0`, because the latter actually performs a `SELECT COUNT…`

jontsai (changes requested):

apps/prelaunch/models.py
goztrk: You are right. Updated.

jontsai (approved):

PR #367, opened 2023-04-01
✨ feat: add support to reversed relations on materialized_property depends_on

goztrk:
`depend_on` parameter of `mp.materialized_property` was only working with forwarded relations.

**Explanation**
Taking below example of model relations:
```python
class Profile(models.Model):
    user = models.OneToOneField(settings.AUTH_USER_MODEL, related_name='profile')

class Asset(models.Model):
    user = models.ForeignKey(settings.AUTH_USER_MODEL, related_name='assets')
    is_active = models.BooleanField(default=True)
```

If we want to create a `materialized_property` in `Profile` class for `Asset` count, it wouldn't work because `User` model has no explicit field definition for asset:

```python
class Profile(models.Model):
    # ...
    @mp.materialized_property(
        models.PositiveIntegerField(blank=True),
        depends_on=['user.assets.is_active', ],  # This wouldn't work before
    )
    active_asset_count(self):
        # ...
```

**The cause of Problem**
If [Django Related Descriptors](https://github.com/django/django/blob/main/django/db/models/fields/related_descriptors.py) checked the relations for Forward descriptors and Reverse descriptors are different.

If relation is forward (The current model has a field definition for related model) the next model can be accessed using `field.rel.to`
If the relation is reverse (The relation definition is actually in the target model) the next model can be accessed using `field.rel.related_model`. `field.rel.to` gives the same current model. Not next model.

### DIFF
--- a/apps/mp/services.py
+++ b/apps/mp/services.py
@@ -1,6 +1,6 @@
 # Python Standard Library Imports
 # Future Imports
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
 
 import contextlib
 import logging
@@ -44,8 +44,8 @@ def materialized_property(
     eventually_consistent=False,
     verify_qs=None,
 ):
-    """Decorator to create a Materialized Property
-    """
+    """Decorator to create a Materialized Property"""
+
     def wrap(f):
         return MaterializedPropertySubstitution(
             f,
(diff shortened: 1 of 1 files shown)


apps/mp/services.py
```diff
@@ -468,18 +568,27 @@ def parse_dependency(model, dependency):
             field = getattr(model, attr_name)
             # for ReverseSingleRelatedObjectDescriptor
             if hasattr(field, 'field'):
+                is_reverse = field.__class__.__name__.startswith('Reverse')
                 field = field.field
-                model = field.rel.to
+                model = field.rel.related_model if is_reverse else field.rel.to
```
goztrk, line 573: # Reviewer Note
Main change

jontsai (approved):

PR #368, opened 2023-04-02
Fix related_name while parsing dependencies on MP

goztrk:
`related_name` was setting wrong name in reversed related name while parsing dependencies.

This PR fixes it.

### DIFF
--- a/apps/mp/services.py
+++ b/apps/mp/services.py
@@ -1,10 +1,7 @@
-# Python Standard Library Imports
 # Future Imports
 from __future__ import absolute_import, print_function
 
-import contextlib
-import logging
-import threading
+# Python Standard Library Imports
 import time
 import traceback
 import weakref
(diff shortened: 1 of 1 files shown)


apps/mp/services.py
```diff
@@ -1,10 +1,7 @@
-# Python Standard Library Imports
 # Future Imports
 from __future__ import absolute_import, print_function
 
-import contextlib
-import logging
-import threading
+# Python Standard Library Imports
```
goztrk, line 4: # Reviewer Note
`isort` changes

apps/mp/services.py
```diff
@@ -571,7 +567,9 @@ def parse_dependency(model, dependency):
                 is_reverse = field.__class__.__name__.startswith('Reverse')
                 field = field.field
                 model = field.rel.related_model if is_reverse else field.rel.to
-                related_name = field.rel.related_name
+                related_name = (
+                    field.name if is_reverse else field.rel.related_name
+                )
```
goztrk, line 572: # Reviewer Note
Main Change!
The `related_name` should be current attribute name in reverse relations. `field.rel.related_name` points to next Model in path.

jontsai (approved):

PR #371, opened 2023-05-16
Log failed stripe handle

aarthi-axim:
In this PR, logged failed stripe event handles in rollbar.
This could help to debug the error in case of event handle failure.

### DIFF
--- a/lib/stripe_lib/views.py
+++ b/lib/stripe_lib/views.py
@@ -1,6 +1,9 @@
 # Python Standard Library Imports
 import json
 
+# Third Party (PyPI) Imports
+import rollbar
+
 # Django Imports
 from django.http import HttpResponse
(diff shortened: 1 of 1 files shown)


lib/stripe_lib/views.py
```diff
@@ -34,26 +37,36 @@ def stripe_webhook_view(request):
...
+        extra_data = {
+            'data': request.body,
+            'error': '{}'.format(e),
+        }
+        rollbar.report_message(
+            'Stripe event: failed',
+            extra_data=extra_data
+        )
 
     response = HttpResponse(status=200)
```
aarthi-axim, line 71: ## Reviewer Note:
Allowed to return response status code as `200` even in case of failure since Stripe is 3rd party tool.

aarthi-axim:
Closing. Yet to analyse more.

PR #375, opened 2023-05-23
Add keyword argument tz

aarthi-axim:
Added `tz` as keyword argument to `fromtimestamp`, a method of `datetime.datetime`.

### DIFF
--- a/utils/datetime_utils.py
+++ b/utils/datetime_utils.py
@@ -73,11 +73,6 @@ def unix_time_to_datetime(timestamp, tzinfo=pytz.utc):
     return dt
 
 
-def unix_time_to_utcdatetime(timestamp):
-    dt = datetime.datetime.utcfromtimestamp(timestamp).replace(tzinfo=pytz.utc)
-    return dt
-
-
 def iso_datetime_to_unix_time(iso):
     """Converts an ISO datetime string to UNIX timestamp
     """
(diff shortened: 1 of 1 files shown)


utils/datetime_utils.py
```diff
@@ -69,7 +69,7 @@ def datetime_to_unix_time(dt):
 
 
 def unix_time_to_datetime(timestamp, tzinfo=pytz.utc):
-    dt = datetime.datetime.fromtimestamp(timestamp).replace(tzinfo=tzinfo)
+    dt = datetime.datetime.fromtimestamp(timestamp, tz=tzinfo)
```
jontsai, line 72: - Can you rebase off of `master` and delete the other `unix_time_to_utcdatetime` ? https://github.com/hacktoolkit/django-htk/blob/master/utils/datetime_utils.py
- Does this work for both Python 2 and 3?
- Was the goal of the function to return a timezone-aware `datetime.datetime` object? If so, I think we still need to keep the `.replace(...)` call

jontsai (changes requested):

utils/datetime_utils.py
aarthi-axim: - Done.
- It works for both Python2 and Python3.
- This itself returns `timezone-aware datetime`. 

Attached images for reference,
py2

py3

jontsai (approved):
LGTM

PR #376, opened 2023-05-31
fix: changelog app problems

goztrk:
- Removed `emoji` PyPI package dependency
- Removed missed `print` function which caused multiple `no such group` outputs
- Converted some strings to unicode because of emoji characters.


apps/changelog/classes/change_log.py
```diff
@@ -63,7 +63,8 @@ def write_changelog(
         buf.append('')
 
         with open(changelog_file_name, 'w') as f:
-            f.write('\n'.join(buf))
+            string_to_write = '\n'.join(buf)
+            f.write(string_to_write.encode('UTF-8'))
```
jontsai, line 67: Please stylize this as `utf-8` (lowercase)

apps/changelog/classes/log_entry.py
```diff
@@ -83,8 +82,7 @@ def release_version(self):
                             sha=m.group('sha'),
                             branch=m.group('branch'),
                         )
-                    except Exception as ex:
-                        print(ex)
+                    except Exception:
```
jontsai, line 85: Not in this PR, but can we also fix a spelling error in `regex.py`

`htk/apps/changelog/constants/regexes.py`
- `# never format: 'tag: deploy-202211031730-4acb099a93-master'`
- Should read `newer` instead of `never`

jontsai (changes requested):

apps/changelog/classes/change_log.py
goztrk: Done

apps/changelog/classes/log_entry.py
goztrk: Done

apps/changelog/classes/log_entry.py
```diff
@@ -3,12 +3,11 @@
 
 # Third Party (PyPI) Imports
 import dateparser
-import emoji
 
 # HTK Imports
 from htk.apps.changelog.classes.release_version import ReleaseVersion
 from htk.apps.changelog.constants.general import SEPARATOR
-from htk.apps.changelog.constants.regexes import (
+from htk.apps.changelog.constants.regexex import (
```
jontsai, line 10: This import looks wrong...

apps/changelog/constants/regexex.py
```diff
@@ -3,7 +3,7 @@
```
jontsai, line 3: Why did this file rename?

apps/changelog/utils.py
```diff
@@ -3,7 +3,7 @@
 
 # HTK Imports
 from htk.apps.changelog.constants.general import SEPARATOR
-from htk.apps.changelog.constants.regexes import ORIGIN_URL_REGEX
+from htk.apps.changelog.constants.regexex import ORIGIN_URL_REGEX
```
jontsai, line 6: Why did this import rename?

jontsai (changes requested):

apps/changelog/constants/regexex.py
goztrk: I understood the previous review wrong and reverted changes.

jontsai (approved):
LGTM

PR #382, opened 2023-07-28
WIP Add Job Information

maxBruckhaus:
WIP to check if this is a feasible method of adding creation time and queue position attributes to the `safe_timed_task` class. Then, can use this information to display in the admin tools.

### DIFF
--- a/decorators/celery_.py
+++ b/decorators/celery_.py
@@ -19,12 +19,14 @@ class safe_timed_task(object):
     def __init__(self, task_name, notify=False):
         self.task_name = task_name
         self.notify = notify
-        self.time_created = datetime.now()
+
+        current_time = datetime.now()
+        formatted_time = current_time.strftime('%Y-%m-%d %H:%M:%S')
+        self.time_created = formatted_time
 
         with celery_app.connection() as connection:
             queue = connection.default_channel.queue_declare(queue='zippy_prod_celery', passive=True)
             queue_position = queue.message_count + 1
-        
         self.queue_position = queue_position
 
     def __call__(self, task_fn):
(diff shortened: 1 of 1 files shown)


decorators/celery_.py
```diff
@@ -1,5 +1,6 @@
 # Python Standard Library Imports
 from functools import wraps
+from datetime import datetime
```
goztrk, line 3: This should be only `import datetime` it is a general rule that we follow.

decorators/celery_.py
```diff
@@ -10,12 +11,24 @@
 from htk.utils.notifications import slack_notify
 from htk.utils.timer import HtkTimer
 
+# Zippy Imports
+from zippy.celeryconfig import app as celery_app
```
goztrk, line 15: Zippy is not an existing module inside `django-htk`.
`django-htk` is a general purpose Tool for Django apps. Whatever we change in htk, we make sure that it works for all projects without depending any Django projects that using `django-htk`

goztrk (changes requested):

decorators/celery_.py
maxBruckhaus: I see, thank you. I will come up with a different way to handle this logic.

decorators/celery_.py
```diff
@@ -10,12 +11,24 @@
...
         self.notify = notify
 
+        current_time = datetime.now()
+        formatted_time = current_time.strftime('%Y-%m-%d %H:%M:%S')
+        self.time_created = formatted_time
+
+        with celery_app.connection() as connection:
+            queue = connection.default_channel.queue_declare(queue='zippy_prod_celery', passive=True)
+            queue_position = queue.message_count + 1
+        self.queue_position = queue_position
```
goztrk, line 30: Instead of this addition, you might want to create a separate class called `safe_timed_tasks_with_queue_position` that extends from `safe_timed_task` class and let it accept `celery_app` in its `__init__` function as a parameter. That way dependency can be provided where it is being used.

PR #384, opened 2023-08-16
✨ feat: Add JavaScript URLs app

goztrk:
Added new template tag to be able to export desired URLs to use in JavaScript.


apps/js_urls/utils.py
```diff
@@ -0,0 +1,18 @@
+# Python Standard Library
+# Python Standard Library Imports
+import typing as t
```
jontsai, line 3: `import typing as T`

apps/js_urls/utils.py
```diff
@@ -0,0 +1,18 @@
+# Python Standard Library
```
jontsai, line 1: Just need one of these labels, not both

apps/js_urls/utils.py
```diff
@@ -0,0 +1,18 @@
...
+# Python Standard Library Imports
+import typing as t
+from functools import reduce
+
+
+def replace(data, replacements: t.List[t.Tuple[str, str]]):
+    """
```
jontsai, line 8: Docstring should not start with newline

apps/js_urls/utils.py
```diff
@@ -0,0 +1,18 @@
...
+    Allows to perform several string substitutions.
+
+    This function performs several string replacements on the initial `data`
+    string using a list of 2-iterables `(old, new)`.
```
jontsai, line 12: `pairs` or `2-tuples` sounds  better

apps/js_urls/README.md
```diff
@@ -0,0 +1,36 @@
+# JavaScript URL Exporter
```
jontsai, line 1: - `js_routes` might sound better?
- Is there any reason this is a separate app under HTK and not just directly in HTK? I'm thinking that apps are better if they also have `models` or substantial amount of functionality.

apps/js_urls/js/build_url.ts
```diff
@@ -0,0 +1,15 @@
+import _ from 'lodash';
```
jontsai, line 1: More modern practice: import `zip` or `flatMap` directly instead of `_`, to minimize the JS footprint.

apps/js_urls/templatetags/js_urls.py
```diff
@@ -0,0 +1,22 @@
...
+from htk.apps.js_urls.serializer import get_urls_as_json
+from htk.utils import htk_setting
+
+
+register = template.Library()
+
+
+@register.simple_tag
+@mark_safe
+def print_js_urls():
```
jontsai, line 15: - Would prefer to put this directly in `htk/tempatetags/**` so that we can just `{% load htk_tags %}`
- I think `print_js_routes`sounds better?

apps/js_urls/templatetags/js_urls.py
```diff
@@ -0,0 +1,22 @@
+# Django Imports
+from django import template
+from django.utils.safestring import mark_safe
+
+# HTK Imports
+from htk.apps.js_urls.serializer import get_urls_as_json
```
jontsai, line 6: Too many functions name `get_*`

apps/js_urls/templatetags/js_urls.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@register.simple_tag
+@mark_safe
+def print_js_urls():
+    if not htk_setting('HTK_APPS_JS_URLS_ENABLED'):
+        return ""
+
+    js_obj = htk_setting('HTK_APPS_JS_URLS_JS_OBJECT_NAME')
+    urls = get_urls_as_json()
```
jontsai, line 20: `build_routes(as_json=True)`

apps/js_urls/serializer.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def get_urls_as_dict(resolver=None):
+    resolver = resolver or get_resolver()
+    return dict(_parse_resolver(resolver))
+
+
+def get_urls_as_json(resolver=None):
+    resolver = resolver or get_resolver()
+    return json.dumps(get_urls_as_dict(resolver))
```
jontsai, line 33:
```suggestion
def build_routes(resolver=None, as_json=False):
    resolver = resolver or get_resolver()
    routes = dict(_parse_resolver(resolver))
    if as_json:
        routes = json.dumps(routes)
        
    return routes
```

apps/js_urls/serializer.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def _parse_resolver(
+    resolver, current_namespace=None, url_prefix=None, include_all=False
+):
+    allowed_urls = htk_setting('HTK_APPS_JS_URLS_ALLOWED')
```
jontsai, line 39: - I think the name should be something like `HTK_JS_ROUTES_URL_NAMES`
- `ALLOWED` doesn't make as much sense because there is nothing to allow/deny.

apps/js_urls/serializer.py
```diff
@@ -0,0 +1,130 @@
...
+    return dict(_parse_resolver(resolver))
+
+
+def get_urls_as_json(resolver=None):
+    resolver = resolver or get_resolver()
+    return json.dumps(get_urls_as_dict(resolver))
+
+
+def _parse_resolver(
+    resolver, current_namespace=None, url_prefix=None, include_all=False
```
jontsai, line 37: Any reason to name the arg `current_namespace` as opposed to just `namespace`?

apps/js_urls/serializer.py
```diff
@@ -0,0 +1,130 @@
...
+):
+    allowed_urls = htk_setting('HTK_APPS_JS_URLS_ALLOWED')
+    namespace = (
+        f"{current_namespace}:{resolver.namespace}"
+        if current_namespace
+        else resolver.namespace
+    )
```
jontsai, line 44: For this pattern, prefer multiple lines instead of a ternary, splitting out the `prefix`. If we also needed to have a suffix, then you could add the `suffix` variable.

```
prefix = f'{current_namespace}:' if current_namespace else ''
namespace = f'{prefix}{resolver.namespace}'
```

apps/js_urls/serializer.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def get_urls_as_json(resolver=None):
+    resolver = resolver or get_resolver()
+    return json.dumps(get_urls_as_dict(resolver))
+
+
+def _parse_resolver(
+    resolver, current_namespace=None, url_prefix=None, include_all=False
+):
```
jontsai, line 38: This function needs a docstring 😄

apps/js_urls/serializer.py
```diff
@@ -0,0 +1,130 @@
...
+    allowed_urls = htk_setting('HTK_APPS_JS_URLS_ALLOWED')
+    namespace = (
+        f"{current_namespace}:{resolver.namespace}"
+        if current_namespace
+        else resolver.namespace
+    )
+    include_all = include_all or (namespace in allowed_urls)
+
+    urls = []
+    for url_pattern in resolver.url_patterns:
```
jontsai, line 48: Not sure what this function is doing as it doesn't have a docstring. The different sections could also use comments.

Unless performing side effects, it might be better to write it as `urls = [_some_transform_fn(url_pattern) for url_pattern in resolver.url_patterns]`

apps/js_urls/constants/regexp.py
```diff
@@ -0,0 +1,9 @@
+# Python Standard Library Imports
+import re
+
+
+URL_ARG_RE = re.compile(r"(\(.*?\))")
```
jontsai, line 5: These patterns are kind of convoluted.

jontsai (changes requested):

apps/js_urls/utils.py
goztrk: Removed typing to make it compatible with Python 2

apps/js_urls/utils.py
goztrk: Fixed

apps/js_urls/utils.py
goztrk: Changed to `2-tuples`

apps/js_urls/README.md
goztrk: Removed as an `app` and added as a templatetag to `htk_tags`.
default constants moved to `htk.constants.defaults`

apps/js_urls/js/build_url.ts
goztrk: This example file is now removed. I couldn't find a good place to put in.

apps/js_urls/templatetags/js_urls.py
goztrk: Done.

apps/js_urls/templatetags/js_urls.py
goztrk: This function renamed to `build_routes` as you suggested.

apps/js_urls/templatetags/js_urls.py
goztrk: Done

apps/js_urls/serializer.py
goztrk: Done

apps/js_urls/serializer.py
goztrk: Done

apps/js_urls/serializer.py
goztrk: This is a recursive function. And Django URL namespaces can be nested.

apps/js_urls/serializer.py
goztrk: I tried this and it is not working because if there is no namespace, `resolver.namespace` is `None` and with using `f` string, it adds as `None` instead of an empty string and becomes something in the screenshot below:

image

apps/js_urls/serializer.py
goztrk: Added. Sorry 😄

apps/js_urls/serializer.py
goztrk: There are two types of url items in `url_patterns` list in `urls.py`

One is normal URL with a view function defined to it like `re_path(r'^index$', views.index_page, name='index')`
Other one is including another `urls.py` with include `re_path(r'^account', include('account.urls'))`

We are traversing nested url patterns with same function.

Added descriptive inline comments for this. I think code can explain better than me 😄

apps/js_urls/serializer.py
```diff
@@ -0,0 +1,130 @@
...
+                else url_pattern.name
+            )
+            if url_name in allowed_urls or include_all:
+                full_url = _prepare_url_part(url_pattern)
+                urls.append((url_name, urljoin(url_prefix or "/", full_url)))
+
+    return urls
+
+
+def _prepare_url_part(url_pattern):
```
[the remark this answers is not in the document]
goztrk: Added

apps/js_urls/constants/regexp.py
goztrk: These are catching regex match groups in URLs like `(\d+)` and named match groups like `(?P<...>...)`

apps/js_urls/utils.py
goztrk: This file contents moved to `htk/utils/strings.py`

utils/strings.py
```diff
@@ -0,0 +1,31 @@
+# Python Standard Library Imports
```
jontsai, line 1: Please move these to https://github.com/hacktoolkit/django-htk/blob/master/utils/text/transformers.py or whichever file is most appropriate in that directory.

utils/strings.py
```diff
@@ -0,0 +1,31 @@
...
+from functools import reduce
+
+
+def replace(data, replacements):
```
jontsai, line 5: - If the initial value is a string, rename this param `s` instead of `data`
- The function could probably go inside `htk/utils/text/general.py`
- Maybe the function signature should be `replace_many(s, replacements)`

jontsai: The function could probably go inside `htk/utils/text/general.py`

constants/defaults.py
```diff
@@ -80,6 +80,15 @@
 # Enums
 HTK_ENUM_SYMBOLIC_NAME_OVERRIDES = {}
 
+##
+# JS Routes
+HTK_JS_ROUTES_ENABLED = False
+HTK_JS_ROUTER_USE_CAMEL_CASE_URL_NAMES = False
```
jontsai, line 86: Should the default be `True`?

constants/defaults.py
```diff
@@ -80,6 +80,15 @@
 # Enums
 HTK_ENUM_SYMBOLIC_NAME_OVERRIDES = {}
 
+##
+# JS Routes
+HTK_JS_ROUTES_ENABLED = False
```
jontsai, line 85: Why does this need to be a setting? It will be fully controlled code. If we don't want to print out any routes, don't call the function in the first place.

Unless you were thinking of always adding this to certain base templates? In which case, I would still argue that we should just add it to templates when we explicitly want to do it.

templatetags/htk_tags.py
```diff
@@ -458,3 +458,21 @@ def credit_card_icon(credit_card_brand):
     else:
         credit_card_icon = DEFAULT_CREDIT_CARD_ICON
     return credit_card_icon
+
+
+@register.simple_tag()
+@mark_safe
+def print_js_routes():
+    if htk_setting('HTK_JS_ROUTES_ENABLED'):
```
jontsai, line 466: Is this guard necessary?

jontsai (changes requested):
- Since we are not making this an `app`, where can we document its usage?
- I'm debating whether it makes sense to introduce the namespace `htk.services`, in which case you could have `htk/services/js_route_serializer/(README.md|utils.py|...)`

templatetags/htk_tags.py
goztrk: Removed.

constants/defaults.py
goztrk: Yes you are correct. It does not makes sense. Removed.

constants/defaults.py
goztrk: Done

templatetags/htk_tags.py
```diff
@@ -458,3 +458,18 @@ def credit_card_icon(credit_card_brand):
...
+
+
+@register.simple_tag()
+@mark_safe
+def print_js_routes():
+    from htk.utils.js_route_serializer import build_routes
+    js_obj = htk_setting('HTK_JS_ROUTES_JS_OBJECT_NAME')
+    urls = build_routes(as_json=True)
+    html = '<script type="text/javascript">'
+    html += '{js_obj} = Object.assign({js_obj} || {{}}, {{ urls: {urls} }});'.format(
```
jontsai, line 470: - Personally, I feel that the `Object.assign(window.Site)` code is going to be very specific.
- Note that `urls` as a key is hardcoded
- I think the `<script type="text/javascript">` is also a bit overextending in reach
- I think the most useful is `build_routes(as_json=True)`
- With these things simplified, would this tag still need `@mark_safe` ?

constants/defaults.py
```diff
@@ -80,6 +80,14 @@
 # Enums
 HTK_ENUM_SYMBOLIC_NAME_OVERRIDES = {}
 
+##
+# JS Routes
+HTK_JS_ROUTER_USE_CAMEL_CASE_URL_NAMES = True
+HTK_JS_ROUTES_URL_NAMES = []
+HTK_JS_ROUTES_DYNAMIC_PART_PLACEHOLDER = '{0}'
+HTK_JS_ROUTES_JS_OBJECT_NAME = 'window.Site'
```
jontsai, line 88: Not sure if this is valuable

jontsai (approved):

PR #387, opened 2023-09-06
✨ feat: Add invitation system to organizations app

goztrk:
# Status
**READY**

# Description
- Added a util method to organizations app to send or re-send invitation email to requested e-mail address
- Added a class based view for extending in project to build up invitation response page


apps/organizations/views.py
```diff
@@ -0,0 +1,86 @@
...
+        token = kwargs.pop('token')
+        OrganizationInvitation = resolve_model_dynamically(
+            htk_setting('HTK_ORGANIZATION_INVITATION_MODEL')
+        )
+        invitation = get_object_or_404(OrganizationInvitation, token=token)
+        is_authenticated = request.user.is_authenticated
+
+        # Logged in user cannot accept invitation in place of someone else.
+        if is_authenticated and request.user.email != invitation.email:
+            raise Http404
```
jontsai, line 54:
```suggestion
        user = request.user if request.user.is_authenticated else None

        if user and user.email != invitation.email and not(user.profile.has_email(email)):
            raise Http404
```

django-htk supports 1:M user:email

See: https://github.com/hacktoolkit/django-htk/blob/e05822a11d01add512ecd6cc42b92a3e28a8bffd/apps/accounts/models.py#L227

apps/organizations/utils.py
```diff
@@ -1,4 +1,13 @@
+# Django Imports
+from django.apps import apps
+from django.contrib.sites.shortcuts import get_current_site
+
 # HTK Imports
+from htk.apps.organizations.emailers import send_invitation_email
+from htk.apps.prelaunch.utils import (
+    PrelaunchSignup,
```
jontsai, line 8: I was previously experiencing a lot of issues with importing `PrelaunchSignup` model if the app is not actually installed. We should probably look into this, or import this at the function level and catch the import error.

jontsai (changes requested):

apps/organizations/views.py
goztrk: Done

apps/organizations/utils.py
goztrk: Done

apps/organizations/utils.py
```diff
@@ -29,3 +35,45 @@ def get_user_organizations_with_attribute(user, key):
...
+
+    Args:
+        request (Request): Request object. It is needed to build full URL and get current site
+        invitation (BaseAbstractOrganizationInvitation): Invitation object can be come from form.
+    """
+    early_access_code = None
+    if (
+        apps.is_installed('htk.apps.prelaunch')
+        and is_prelaunch_mode()
+        and htk_setting('HTK_ORGANIZATION_INVITATION_AUTO_ADD_TO_WAIT_LIST')
```
jontsai, line 56: Is this setting necessary? If it were `False`, how would this even work?

apps/organizations/utils.py
```diff
@@ -29,3 +35,45 @@ def get_user_organizations_with_attribute(user, key):
         organization__attributes__value=None
     )
     return organizations
+
+
+def invite_person_to_organization(request, invitation):
```
jontsai, line 40: Rename to `invite_organization_member`

apps/organizations/utils.py
```diff
@@ -29,3 +35,45 @@ def get_user_organizations_with_attribute(user, key):
...
+    if invitation.accepted is not None:
+        # If invitation is not Invited status, update to Invited.
+        invitation.accepted = None
+    else:
+        pass
+
+    # Save invitation so that updatedAt field will be updated
+    invitation.save()
+
+    send_invitation_email(request, invitation, early_access_code)
```
jontsai, line 79:
```suggestion
    send_invitation_email(request, invitation, early_access_code=early_access_code)
```

Use the named arg when it is calling the function that has a default arg

apps/organizations/views.py
```diff
@@ -0,0 +1,95 @@
...
+from django.views import View
+
+# HTK Imports
+from htk.apps.organizations.enums import OrganizationMemberRoles
+from htk.utils import (
+    htk_setting,
+    resolve_model_dynamically,
+)
+from htk.view_helpers import render_custom as _r
+from htk.view_helpers import wrap_data
```
jontsai, line 16: I think there is a way to get a dynamic `wrap_data` function from `settings.py`

apps/organizations/views.py
```diff
@@ -0,0 +1,95 @@
...
+
+        self.invitation.save()
+
+        self.data['invitation_accepted'] = self.invitation.accepted
+
+        response = self.render_method(request, self.template_name, self.data)
+        return response
+
+    def build_context(self):
+        self.data = wrap_data(self.request)
```
jontsai, line 95: Use the dynamic `wrap_data` not the generic one

jontsai (changes requested):

apps/organizations/utils.py
goztrk: If it is `False` it will not append any `early_access_code` to invitation. I might help to prevent the people in the Team which is not an engineer and accidentally inviting someone before it is ready while testing. Maybe it is a long shot?

apps/organizations/constants/defaults.py
```diff
@@ -27,3 +27,8 @@
     'user__last_name',
     'user__username',
 )
+
+HTK_ORGANIZATION_INVITATION_RESPONSE_URL_NAME = ''
+HTK_ORGANIZATION_INVITATION_EMAIL_TEMPLATE_NAME = ''
+HTK_ORGANIZATION_INVITATION_EMAIL_SUBJECT = 'You have been invited to join {}'
+HTK_ORGANIZATION_INVITATION_AUTO_ADD_TO_WAIT_LIST = True
```
jontsai, line 34: I think this should just always be true, unless you think otherwise?

And if it's always true, let's just remove the setting and condition in the code.

apps/organizations/utils.py
jontsai: Yeah, IMO, not necessary, just remove

apps/organizations/views.py
```diff
@@ -0,0 +1,98 @@
...
+        self.invitation.save()
+
+        self.data['invitation_accepted'] = self.invitation.accepted
+
+        response = self.render_method(request, self.template_name, self.data)
+        return response
+
+    def build_context(self):
+        wrap_data = resolve_method_dynamically(
+            htk_setting('HTK_PRELAUNCH_VIEW_CONTEXT_GENERATOR')
```
jontsai, line 96: Can you change this to `HTK_VIEW_CONTEXT_GENERATOR`, as well as in `apps/prelaunch/**` ?

jontsai (changes requested):

apps/organizations/utils.py
goztrk: Done.

apps/organizations/utils.py
goztrk: Done

apps/organizations/utils.py
goztrk: Done

apps/organizations/views.py
goztrk: Done used `HTK_VIEW_CONTEXT_GENERATOR` setting

apps/organizations/views.py
goztrk: Done used `HTK_VIEW_CONTEXT_GENERATOR` setting

apps/organizations/constants/defaults.py
goztrk: Removed

apps/organizations/views.py
goztrk: Done

jontsai (approved):
LGTM

PR #392, opened 2023-10-12
feat: ⚡️ Improve MP dependency parser for Django >=2

goztrk:
`htk.apps.mp.services.parse_dependency` function is updated to work correctly on Django versions 2.x, 3.x and 4.x

### DIFF
--- a/apps/mp/services.py
+++ b/apps/mp/services.py
@@ -578,7 +578,7 @@ def parse_dependency(model, dependency):
                 else:
                     model = field.related.model
                     related_name = field.related.field.name
-            else:  # Django version is >= 2 - (Tested on Django 2.2.28, 3.2.22, 4.2.6)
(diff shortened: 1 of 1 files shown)


apps/mp/services.py
```diff
@@ -562,24 +563,53 @@ def parse_dependency(model, dependency):
...
 
-                if not related_name or related_name == '+':
-                    raise Exception(
-                        "Fields with no explicit related name aren't supported for now: {}".format(
-                            field
-                        )
+                else:
+                    model = field.related.model
+                    related_name = field.related.field.name
+            else:  # Django version is >= 2
```
jontsai, line 581: What versions of Django has this been tested with? Can we leave some kind of comment indicating the level of confidence for this piece of code being compatible with Django 2, and Django >= 3?

PR #393, opened 2023-10-19
update OrganizationInvitation signal handler for sending Slack messages

lleung030:
Updates the signal handler for `OrganizationInvitation` to send a Slack notification when an invitation is created, re-sent, accepted, or declined.

### DIFF
--- a/apps/organizations/apps.py
+++ b/apps/organizations/apps.py
@@ -31,6 +31,7 @@ def organization_invitation_created_or_updated(sender, instance, created, **kwar
             False: invitation.build_notification_message__declined,
             None: invitation.build_notification_message__resent,
         }
+
         msg_builder_key = 'created' if created else invitation.accepted
         msg_builder = INVITATION_ACCEPTANCE_MESSAGE_BUILDERS[msg_builder_key]
         msg = msg_builder()
--- a/apps/organizations/apps.py
+++ b/apps/organizations/apps.py
@@ -35,6 +35,7 @@ def organization_invitation_created_or_updated(sender, instance, created, **kwar
         msg_builder_key = 'created' if created else invitation.accepted
         msg_builder = INVITATION_ACCEPTANCE_MESSAGE_BUILDERS[msg_builder_key]
         msg = msg_builder()
+        
         try:
             slack_notify(msg)
         except Exception:
--- a/apps/organizations/apps.py
+++ b/apps/organizations/apps.py
@@ -19,48 +19,28 @@ from htk.utils import (
 
 
 @disable_for_loaddata
-def organization_invitation_created(sender, instance, created, **kwargs):
+def organization_invitation_updated(sender, instance, created, **kwargs):
     """Signal handler for when a new OrganizationInvitation object is created or updated"""
-    if created:
-        invitation = instance
+    invitation = instance
+    if not settings.TEST and htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
+        from htk.utils.notifications import slack_notify
 
-        if not settings.TEST and htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
+        if created:
+            msg = invitation.build_notification_message__created()
+        else:
+            INVITATION_ACCEPTANCE_MESSAGE_BUILDERS = {
+                True: invitation.build_notification_message__accepted,
+                False: invitation.build_notification_message__declined,
+                None: invitation.build_notification_message__resent,
+            }
+            msg_builder = INVITATION_ACCEPTANCE_MESSAGE_BUILDERS.get(invitation.accepted)
+
+        if msg_builder is not None:
+            msg = msg_builder()
             try:
-                from htk.utils.notifications import slack_notify
-
-                msg = invitation.build_notification_message__created()
                 slack_notify(msg)
-            except:
+            except Exception:
                 rollbar.report_exc_info()
-    if created == False:
-        invitation = instance
-
-        if not settings.TEST and htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
-            if invitation.accepted == True:
-                try:
-                    from htk.utils.notifications import slack_notify
-
-                    msg = invitation.build_notification_message__accepted()
-                    slack_notify(msg)
-                except:
-                    rollbar.report_exc_info()
-            if invitation.accepted == False:
-                try:
-                    from htk.utils.notifications import slack_notify
-
-                    msg = invitation.build_notification_message__declined()
-                    slack_notify(msg)
-                except:
-                    rollbar.report_exc_info()
-            if invitation.accepted == True:
-                try:
-                    from htk.utils.notifications import slack_notify
-
-                    msg = invitation.build_notification_message__resent()
-                    slack_notify(msg)
-                except:
-                    rollbar.report_exc_info()
-
 
 class HtkOrganizationAppConfig(HtkAppConfig):
     name = 'htk.apps.organizations'
--- a/apps/organizations/apps.py
+++ b/apps/organizations/apps.py
@@ -20,7 +20,7 @@ from htk.utils import (
 
 @disable_for_loaddata
 def organization_invitation_created(sender, instance, created, **kwargs):
-    """Signal handler for when a new OrganizationInvitation object is created"""
+    """Signal handler for when a new OrganizationInvitation object is created or updated"""
     if created:
         invitation = instance
 
@@ -32,6 +32,34 @@ def organization_invitation_created(sender, instance, created, **kwargs):
                 slack_notify(msg)
             except:
                 rollbar.report_exc_info()
+    if created == False:
+        invitation = instance
+
+        if not settings.TEST and htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
+            if invitation.accepted == True:
+                try:
+                    from htk.utils.notifications import slack_notify
+
+                    msg = invitation.build_notification_message__accepted()
+                    slack_notify(msg)
+                except:
+                    rollbar.report_exc_info()
+            if invitation.accepted == False:
+                try:
+                    from htk.utils.notifications import slack_notify
+
+                    msg = invitation.build_notification_message__declined()
+                    slack_notify(msg)
+                except:
+                    rollbar.report_exc_info()
+            if invitation.accepted == True:
+                try:
+                    from htk.utils.notifications import slack_notify
+
+                    msg = invitation.build_notification_message__resent()
+                    slack_notify(msg)
+                except:
+                    rollbar.report_exc_info()
 
 
 class HtkOrganizationAppConfig(HtkAppConfig):
--- a/apps/organizations/apps.py
+++ b/apps/organizations/apps.py
@@ -19,28 +19,25 @@ from htk.utils import (
 
 
 @disable_for_loaddata
-def organization_invitation_updated(sender, instance, created, **kwargs):
+def organization_invitation_created_or_updated(sender, instance, created, **kwargs):
     """Signal handler for when a new OrganizationInvitation object is created or updated"""
     invitation = instance
     if not settings.TEST and htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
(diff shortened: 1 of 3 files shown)


apps/organizations/apps.py
```diff
@@ -32,6 +32,34 @@ def organization_invitation_created(sender, instance, created, **kwargs):
                 slack_notify(msg)
             except:
                 rollbar.report_exc_info()
+    if created == False:
```
jontsai, line 35: In Python, do not compare booleans using `==`. Compare using the `is` operator.

For more on this, see: https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is

Also, you can/should just handle this using the `else` clause.

apps/organizations/apps.py
```diff
@@ -32,6 +32,34 @@ def organization_invitation_created(sender, instance, created, **kwargs):
...
+        invitation = instance
+
+        if not settings.TEST and htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
```
jontsai, line 38: - Lots of repetition here.
- Could/should use the `elif` clause.

apps/organizations/apps.py
```diff
@@ -32,6 +32,34 @@ def organization_invitation_created(sender, instance, created, **kwargs):
...
+                except:
+                    rollbar.report_exc_info()
+            if invitation.accepted == True:
+                try:
+                    from htk.utils.notifications import slack_notify
+
+                    msg = invitation.build_notification_message__resent()
+                    slack_notify(msg)
+                except:
+                    rollbar.report_exc_info()
```
jontsai, line 62: - Generally, you should not be re-defining or re-importing the same variables multiple times. The common ones can go outside of the `if/elif/else blocks`, and only the specific code can go into each block.
- For `accepted`, since it is a Null BooleanField, we do have to compare `is True`, `is False`, or `is None` since it can have 3 values instead of the normal 2 (`True`, `False`) for `BooleanField

Here's how I would clean up this function:

```
@disable_for_loaddata
def organization_invitation_updated(sender, instance, created, **kwargs):
    """Signal handler for when a new OrganizationInvitation object is created or updated"""
    invitation = instance
    if not settings.TEST and htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
        from htk.utils.notifications import slack_notify

        if created:
            msg = invitation.build_notification_message__created()
        else:
            if invitation.accepted is True:
                    msg = invitation.build_notification_message__accepted()
            elif invitation.accepted is False
                    msg = invitation.build_notification_message__declined()
            elif invitation.accepted is None
                    msg = invitation.build_notification_message__resent()
            else:
                # impossible case
                msg = None

        if msg:
            try:
                slack_notify(msg)
            except:
                rollbar.report_exc_info()
```

Then, this cascading if-elif-else can be further cleaned up using another "clever trick" Python pattern:

```
@disable_for_loaddata
def organization_invitation_updated(sender, instance, created, **kwargs):
    """Signal handler for when a new OrganizationInvitation object is created or updated"""
    invitation = instance
    if not settings.TEST and htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
        from htk.utils.notifications import slack_notify

        if created:
            msg = invitation.build_notification_message__created()
        else:
            INVITATION_ACCEPTANCE_MESSAGE_BUILDERS = {
                True: invitation.build_notification_message__accepted,
                False: invitation.build_notification_message__declined,
                None: invitation.build_notification_message__resent,
            }
            msg_builder = INVITATION_ACCEPTANCE_MESSAGE_BUILDERS.get(invitation.accepted)

        if msg_builder is not None:
            msg = msg_builder()
            try:
                slack_notify(msg)
            except Exception:
                rollbar.report_exc_info()
```

Note on exception handling:
- We want to put as little code as possible in the `try` block, and only wrap the critical section of code which can potentially fail inside of the `try` block
- Always catch *something*. Stylistically, it's good to write `except Exception` rather than just `except`.

jontsai (changes requested):
You will also need to implement the base functions in https://github.com/hacktoolkit/django-htk/blob/b469817e147824c1b5dfbc96709e5f6fca56abbf/apps/organizations/models.py#L320-L333

apps/organizations/apps.py
lleung030, line 1: Screenshot 2023-10-19 at 4 49 59 PM

Although the function is able to send a signal, the parameters still need to addressed in https://github.com/aweandreverence/maskil-django/pull/27

jontsai (changes requested):

apps/organizations/models.py
```diff
@@ -320,18 +320,47 @@ def status(self) -> str:
...
     # Notifications
 
+    def _build_notification_message(self, subject: User, verb: str) -> str:
+        """Builds a message that will be displayed as an internal Slack notification.
+        """
+        msg = '{subject_name} ({subject_username}<{subject_profile_url}|{email}>) has {verb} an invitation for Organization <{organization_url}|{organization_name}>'.format( # noqa: E501
+            verb=verb,
+            subject_name=subject.profile.get_full_name(),
+            subject_username=subject.username,
+            subject_profile_url=subject.profile.profile_url,
```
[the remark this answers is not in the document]
lleung030: I understand now. Thanks!

lleung030 (commented):
I've reviewed it and it looks complete.

apps/organizations/apps.py
```diff
@@ -19,21 +19,29 @@
...
+            }
+            msg_builder = INVITATION_ACCEPTANCE_MESSAGE_BUILDERS.get(invitation.accepted)
+
+        if msg_builder is not None:
+            msg = msg_builder()
             try:
-                from htk.utils.notifications import slack_notify
-
-                msg = invitation.build_notification_message__created()
                 slack_notify(msg)
```
jontsai, line 41: Unless you ensure that all of the functions will always return a message, you will also need to check `if msg is not None` because it looks like there is a case where you don't return a message.

apps/organizations/models.py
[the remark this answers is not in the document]
jontsai: This needs to be removed...

apps/organizations/models.py
```diff
@@ -320,18 +320,47 @@ def status(self) -> str:
     ##
     # Notifications
 
+    def _build_notification_message(self, subject: User, verb: str) -> str:
+        """Builds a message that will be displayed as an internal Slack notification.
+        """
+        msg = '{subject_name} ({subject_username}<{subject_profile_url}|{email}>) has {verb} an invitation for Organization <{organization_url}|{organization_name}>'.format( # noqa: E501
```
jontsai, line 326: You have it inverted.

jontsai (changes requested):

apps/organizations/views.py
```diff
@@ -77,8 +77,8 @@ def post(self, request, *args, **kwargs):
         self.invitation.accepted = invitation_response == 'accept'
         self.invitation.responded_at = timezone.datetime.now()
 
+        self.invitation.user = request.user
```
jontsai, line 80: 👍

apps/organizations/models.py
```diff
@@ -320,18 +321,48 @@ def status(self) -> str:
     ##
     # Notifications
 
+    User = get_user_model()
```
jontsai, line 324: Remove; you cannot do this here.

apps/organizations/models.py
```diff
@@ -320,18 +321,48 @@ def status(self) -> str:
...
+
+    def _build_notification_message(self, subject: User, verb: str) -> str:
```
jontsai, line 326: For now, HTK has to be Python2 compatible (there are many users of django-htk still on Python 2, sadly), so please remove all of the type-hints as shown in the suggested code

```suggestion
    def _build_notification_message(self, subject, verb):
```

apps/organizations/models.py
```diff
@@ -320,18 +321,48 @@ def status(self) -> str:
...
+        return msg
+
     def build_notification_message__created(self):
         """Builds a message that will be displayed as an internal Slack notification
         when this invitation object is created.
         """
-        msg = '{invited_by_name} ({invited_by_username}) has sent an invitation for organization {organization_name} to {email}'.format(
-            invited_by_name=self.invited_by.profile.get_full_name(),
-            invited_by_username=self.invited_by.username,
-            organization_name=self.organization.name,
```
jontsai, line 330: The original (previous) code only displayed `organization.name`, no `organization.profile_url` because the `BaseAbstractionOrganization` (defined here: https://github.com/aweandreverence/django-htk/blob/b469817e147824c1b5dfbc96709e5f6fca56abbf/apps/organizations/models.py#L61) does not have a `profile_url` attribute.

apps/organizations/models.py
```diff
@@ -320,18 +320,47 @@ def status(self) -> str:
...
+    def _build_notification_message(self, subject: User, verb: str) -> str:
+        """Builds a message that will be displayed as an internal Slack notification.
+        """
+        msg = '{subject_name} ({subject_username}<{subject_profile_url}|{email}>) has {verb} an invitation for Organization <{organization_url}|{organization_name}>'.format( # noqa: E501
+            verb=verb,
+            subject_name=subject.profile.get_full_name(),
+            subject_username=subject.username,
+            subject_profile_url=subject.profile.profile_url,
+            email=subject.email,
+            organization_url=self.organization.profile_url,
```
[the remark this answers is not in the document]
jontsai: Remove this, because there is no such attribute on this class. It _might_ exist on a descendant, but it is not on this (ancestor) class -- that's where you might be getting confused.

jontsai (changes requested):

apps/organizations/apps.py
```diff
@@ -19,21 +19,29 @@
...
+        else:
+            INVITATION_ACCEPTANCE_MESSAGE_BUILDERS = {
+                True: invitation.build_notification_message__accepted,
+                False: invitation.build_notification_message__declined,
+                None: invitation.build_notification_message__resent,
+            }
+            msg_builder = INVITATION_ACCEPTANCE_MESSAGE_BUILDERS.get(invitation.accepted)
+
+        if msg_builder is not None:
+            msg = msg_builder()
```
jontsai, line 39:
```suggestion
        INVITATION_ACCEPTANCE_MESSAGE_BUILDERS = {
            'created': invitation.build_notification_message__created,
            True: invitation.build_notification_message__accepted,
            False: invitation.build_notification_message__declined,
            None: invitation.build_notification_message__resent,
        }

        msg_builder_key = 'created' if created else invitation.accepted
        msg_builder = INVITATION_ACCEPTANCE_MESSAGE_BUILDERS[msg_builder_key]
        msg = msg_builder()
```

jontsai (changes requested):

apps/organizations/apps.py
```diff
@@ -19,20 +19,25 @@
...
+        INVITATION_ACCEPTANCE_MESSAGE_BUILDERS = {
+            'created': invitation.build_notification_message__created,
+            True: invitation.build_notification_message__accepted,
+            False: invitation.build_notification_message__declined,
+            None: invitation.build_notification_message__resent,
+        }
+        msg_builder_key = 'created' if created else invitation.accepted
+        msg_builder = INVITATION_ACCEPTANCE_MESSAGE_BUILDERS[msg_builder_key]
+        msg = msg_builder()
+        try:
```
jontsai, line 37: Add a newline above this.

apps/organizations/models.py
```diff
@@ -320,18 +321,45 @@ def status(self) -> str:
...
         return msg
 
+    def build_notification_message__declined(self):
+        """Builds a message that will be displayed as an internal Slack notification
+        when this invitation object is declined.
+        """
+        msg = self._build_notification_message(self.user, 'declined')
+        return msg
+    
+    def build_notification_message__resent(self):
```
jontsai, line 357: - Put `__resent` function right under `__created`
- Let's get rid of all the docstrings on the `build_notification_message__*` functions -- it's too wordy, for not enough information. The function names are already sufficient in this case.

See: https://en.wikipedia.org/wiki/Self-documenting_code

jontsai (changes requested):

jontsai (approved):

PR #395, opened 2023-10-24
feat(API): ✨ Add get_object_or_json_error util function

goztrk:
## Status
**READY**

## Description
Added `get_object_or_json_error` util function to prevent boilerplate code.

## Motivation & Context
Due to the incapability of `django.shortcuts.get_object_or_404` being used in API Views, it was necessary to write boilerplate code like this:

```python
try:
    user = User.objects.get(id=user_id)
except User.DoesNotExist:
    user = None

if user is None:
    response = json_response_error()
else:
    # Deal with `user` object
```

Thanks to this function it is simplified to:

```python
user = get_object_or_json_error(User, id=user_id)

# Deal with `user` object
```


api/utils.py
```diff
@@ -14,6 +14,7 @@
     HttpResponse,
     QueryDict,
 )
+from django.shortcuts import _get_queryset
```
jontsai, line 17: The `_` in the name suggests that this is a private function. Do you feel like this is stable enough that we won’t get rug-pulled in future versions?

Is there a public function that we could use instead?

api/utils.py
```diff
@@ -162,3 +163,41 @@ def extract_post_params(
...
+
+
+def get_object_or_json_error(klass, *args, **kwargs):
+    """
+    This is exact replica of `django.shortcuts.get_object_or_404()` function
+    but instead the raised error is a `json_response_error()`.
+
+    Default message is `Not Found` and default HTTP status is `404`.
+
+    _status: int = Status code. Default 404
```
jontsai, line 175: Please opt for more words instead of `_`

Change to `status_code` instead

api/utils.py
```diff
@@ -162,3 +163,41 @@ def extract_post_params(
...
+            if isinstance(klass, type)
+            else klass.__class__.__name__
+        )
+        raise ValueError(
+            "First argument to get_object_or_json_error() must be a Model "
+            "Manager or QuerySet, not '%s'." % klass__name
+        )
+
+    error_message = kwargs.pop('_error_message', 'Not Found')
+    status = kwargs.pop('_status', 404)
```
jontsai, line 194:
```suggestion
    status_code = kwargs.pop('status_code', 404)
```

api/utils.py
```diff
@@ -162,3 +163,41 @@ def extract_post_params(
...
+    try:
+        return queryset.get(*args, **kwargs)
+    except queryset.model.DoesNotExist:
+        raise json_response_error(
+            {
+                'message': error_message,
+            },
+            status=status
```
jontsai, line 202:
```suggestion
            status=status_code
```

Only use “their” name until the last minute possible. Up until then, use “our” name when referencing to an object.

jontsai (changes requested):

api/utils.py
goztrk: I copied exact contents of `get_object_or_404` function and it was using this method to get the queryset. I only changed the raised response upon failure. Since this acts like `get_object_or_404`, I think it is okay to use an internal function of Django in this case.

I also checked up to date version and older versions of this function and I think this function stability level is 5/5.

https://github.com/django/django/blob/d22ba0763086f71fb6df8502162ea63944d166b1/django/shortcuts.py#L51

api/utils.py
goztrk: The reason I put underscore prefix was to ensure that it will not conflict with any model field in the future. If you think we won't ever use `status_code` in any Model than it is OK.

goztrk: changed to `http_status_code` to be more specific

api/utils.py
goztrk: changed to `http_status_code` to be more specific

api/utils.py
goztrk: changed to `http_status_code` to be more specific

utils/http/response.py
```diff
@@ -4,3 +4,12 @@
 
 class HttpResponseAccepted(HttpResponse):
     status_code = 202
+
+
+class ResponseError(Exception):
```
jontsai, line 9: Might be better in `utils/http/errors.py` ?

middleware/classes.py
```diff
@@ -130,3 +131,18 @@ def process_request(self, request):
             django_timezone = None
         if django_timezone:
             timezone.activate(django_timezone)
+
+
+class RaiseResponse:
+    """
+    Processes exceptions, returning a response if it's the case or raising
```
jontsai, line 138: This comment is unclear. If "what" is the case?

Also, please start the docstring right after the quotes (don't include an empty newline)

api/utils.py
```diff
@@ -162,3 +168,43 @@ def extract_post_params(
             if value is not None:
                 data[param] = value
     return data
+
+
+def get_object_or_json_error(klass, *args, **kwargs):
+    """
+    This is exact replica of `django.shortcuts.get_object_or_404()` function
```
jontsai, line 175: Please include a permalink to the Django source as part of the docstring. You can add a "References" section toward the end with a bulleted list of links.

api/utils.py
```diff
@@ -162,3 +168,43 @@ def extract_post_params(
...
+    but instead the raised error is a `json_response_error()`.
+
+    Default message is `Not Found` and default HTTP status is `404`.
+
+    http_status_code: int = Status code. Default 404
+    http_error_message: str = Error message. Default `Not Found`
+
+    IMPORTANT NOTE: `htk.middleware.classes.RaiseResponse` MUST be added to
+                    middlewares in settings to make this work!
```
jontsai, line 184: - properly capitalize `middlewares` to whatever the Django variable should be
- rename `settings` to `Django settings`
- Replace the `!` with a `.`

middleware/classes.py
```diff
@@ -130,3 +131,18 @@ def process_request(self, request):
             django_timezone = None
         if django_timezone:
             timezone.activate(django_timezone)
+
+
+class RaiseResponse:
```
jontsai, line 136: I feel like we could come up with a better name

jontsai (changes requested):

utils/http/response.py
goztrk: Done.

middleware/classes.py
goztrk: Updated docstring formatting and changed contents to be more descriptive.

api/utils.py
goztrk: Done.

api/utils.py
goztrk: Done

middleware/classes.py
goztrk: Changed name to `CatchRaisedExceptionResponseMiddleware`

api/utils.py
```diff
@@ -162,3 +168,45 @@ def extract_post_params(
             if value is not None:
                 data[param] = value
     return data
+
+
+def get_object_or_json_error(klass, *args, **kwargs):
```
jontsai, line 173: What if instead we did:

```
def get_object_or_json_error(*args, **kwargs):
    try:
        obj = get_object_or_404(*args, **kwargs)
    except Http404:
          raise ResponseError(
            json_response_error(
                {
                    'message': 'Not Found',
                },
                status=404,
            )
        )

    return obj

```

jontsai (changes requested):

utils/http/errors.py
```diff
@@ -0,0 +1,12 @@
+class ResponseError(Exception):
```
jontsai, line 1: Rename to `HttpErrorResponseError` (I know, two `Error`s in the name -- the first is to denote that there is an HTTP error; the second is to denote that it is a special Python object which extends `Error`/`Exception`)

utils/http/errors.py
```diff
@@ -0,0 +1,12 @@
...
+    """Generic ResponseError exception
```
jontsai, line 2: Let's make the docstring better and more accurate after this next round of changes.

utils/http/errors.py
```diff
@@ -0,0 +1,12 @@
...
+
+    It helps to stop processing a view function by raising an exception.
+    It is like `Http404` exception but with ability to define your response.
+
+    NOTE: `htk.middleware.classes.CatchRaisedExceptionResponseMiddleware` MUST
+          be in MIDDLEWARES in Django Settings.
+    """
+
+    def __init__(self, response):
+        self.response = response
```
jontsai, line 12: I think it should have `response` as well as `status_code` as a parameter.

middleware/classes.py
```diff
@@ -119,14 +117,30 @@ def _is_user_agent_msie(self, request):
...
-            #request.session[DJANGO_TIMEZONE] = django_timezone
+            # request.session[DJANGO_TIMEZONE] = django_timezone
         else:
             django_timezone = None
         if django_timezone:
             timezone.activate(django_timezone)
+
+
+class CatchRaisedExceptionResponseMiddleware:
+    """Catch Raised Exception Response Middleware
```
jontsai, line 134: Change docstring to something like:

```
"""This middleware allows `HttpErrorResponseError`s to be thrown
to short-circuit processing of a view, and allows a response to be returned instead.
"""
```

jontsai (changes requested):

api/utils.py
```diff
@@ -188,3 +196,25 @@ def extract_post_params(
             if value is not None:
                 data[param] = value
     return data
+
+
+def get_object_or_json_error(*args, **kwargs):
+    """
```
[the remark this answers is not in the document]
goztrk: Fixed.

middleware/classes.py
```diff
@@ -119,14 +117,30 @@ def _is_user_agent_msie(self, request):
...
             # <DstTzInfo 'America/Los_Angeles' PST-1 day, 16:00:00 STD> is not JSON serializable
-            #request.session[DJANGO_TIMEZONE] = django_timezone
+            # request.session[DJANGO_TIMEZONE] = django_timezone
         else:
             django_timezone = None
         if django_timezone:
             timezone.activate(django_timezone)
+
+
+class CatchRaisedExceptionResponseMiddleware:
```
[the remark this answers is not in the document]
goztrk: Done

utils/http/errors.py
goztrk: Done

utils/http/errors.py
goztrk: I added a small extra line for `status_code`. If you want any other changes I can add them too.

utils/http/errors.py
goztrk: Done

jontsai: Does `response` always have a `status_code` set or not?

utils/http/errors.py
```diff
@@ -0,0 +1,17 @@
...
+    Status code defaults to 404 but can be overridden.
+
+    NOTE: `htk.middleware.classes.HttpErrorResponseMiddleware` MUST be in
+    MIDDLEWARES in Django Settings.
+    """
+
+    def __init__(self, response, status_code=None):
+        self.response = response
+        self.response.status_code = (
+            status_code if status_code is not None else 404
```
jontsai, line 16: I may have been wrong on this, depending on the answer to my question above.

In any case, if this code is correct locally, then the preference would be to change `404` to be the default value of `status_code` on the function parameters, and change this to just be `self.response.status_code = status_code`

jontsai (changes requested):

utils/http/errors.py
goztrk: `response` must always have a `status_code` and it defaults to `200` if not set.

I changed structure to accept also `string` and update status code if it is `200` because it does not make sense to have status code 200 if this is an raised response error.

utils/http/errors.py
goztrk: Updated this code. You might want to take another look at this. I think it is better now.

utils/http/errors.py
```diff
@@ -0,0 +1,34 @@
...
+    NOTE: `htk.middleware.classes.HttpErrorResponseMiddleware` MUST be in
+    MIDDLEWARES in Django Settings.
+    """
+
+    def __init__(self, response, status_code=None):
+        if isinstance(response, str):
+            self.response = HttpResponse(response)
+            self.response.status_code = (
+                status_code if status_code is not None else 404
+            )
```
jontsai, line 23:
```suggestion
    def __init__(self, response, status_code=400):
        if isinstance(response, str):
            self.response = HttpResponse(response)
            self.response.status_code = status_code
```

- Why not use a default arg?
- Prefer `400` over `404`

Actually, can are you able to set the status code at the time of building, something like `HttpResponse(response, status_code=status_code)` ?

utils/http/errors.py
```diff
@@ -0,0 +1,34 @@
...
+        elif issubclass(response.__class__, HttpResponse):
+            self.response = response
+            # It does not make sense to have status_code 200 for an error
+            if self.response.status_code == 200:
+                self.response.status_code = (
+                    status_code if status_code is not None else 404
+                )
```
jontsai, line 30:
```suggestion
            # replace a 200 status_code with 4xx for errors
            if response.status_code == 200:
                response.status_code = status_code

            self.response = response
```

jontsai (changes requested):

utils/http/errors.py
goztrk: - Used default arg
- Used 400 instead of 404 as default

Yes, the arg name is `status` and used it.

utils/http/errors.py
goztrk: Done. Updated.

jontsai (changes requested):

jontsai (approved):
LGTM

PR #396, opened 2023-10-26
Implement Release Notes support for the Awesome.Bible

lleung030:
Edit `django-htk` files


jontsai:
This is not correct. You can customize this inside of your settings file.

This file is for the defaults for HTK.

PR #400, opened 2023-11-13
feat(Organization): ✨ Add  new API endpoint views to Organization app

goztrk:
## Status
**WIP**

## Description
- Added new API endpoints to deal with
  - Organization Members
  - Invitations
  - Teams
  - Team Members

**NOTE:** It is in the testing phase.

### DIFF
--- a/apps/organizations/decorators.py
+++ b/apps/organizations/decorators.py
@@ -6,10 +6,12 @@ from django.http import (
     Http404,
     HttpResponseForbidden,
 )
-from django.shortcuts import get_object_or_404
 
 # HTK Imports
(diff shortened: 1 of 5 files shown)


apps/organizations/decorators.py
```diff
@@ -122,3 +124,84 @@ def __call__(self, view_func):
         return require_organization_permission(
             OrganizationMemberRoles.MEMBER, content_type=self.content_type
         )(view_func)
+
+
+class require_organization_member_user(object):
```
jontsai, line 129: Not sure if I like `require_` as part of the decorator name prefix for all of these

PR #402, opened 2023-11-19
feat(decorator): ✨ Resolve records from URL

goztrk:
## Status
**READY**

## Description
- Added `resolve_records_from_restful_url` decorator which tries to find the records from path arguments in a chained way.
- Adds `pascal_case_to_snake_case` text transformer util.

### DIFF
--- a/api/utils.py
+++ b/api/utils.py
@@ -23,7 +23,6 @@ from htk.api.constants import (
     HTK_API_JSON_VALUE_ERROR,
     HTK_API_JSON_VALUE_OKAY,
 )
-from htk.models import HtkBaseModel
 from htk.utils.http.errors import HttpErrorResponseError
 
 
@@ -32,6 +31,7 @@ from htk.utils.http.errors import HttpErrorResponseError
 
 class HtkJSONEncoder(serializers.json.DjangoJSONEncoder):
     def default(self, obj):
+        from htk.models import HtkBaseModel
         from django.contrib.auth import get_user_model
 
         UserModel = get_user_model()
--- a/decorators/classes.py
+++ b/decorators/classes.py
@@ -151,7 +151,7 @@ class resolve_records_from_restful_url(object):
                     )
             else:
                 raise ValueError(
-                    f'[{self._get_class_name()}] Invalid model map item type: {type(entry)}'
+                    f'[{self._get_class_name()}] Invalid model map entry type: {type(entry)}'
                 )
 
     def __call__(self, view_fn):
--- a/decorators/classes.py
+++ b/decorators/classes.py
@@ -133,8 +133,8 @@ class resolve_records_from_restful_url(object):
             if isinstance(entry, tuple):
                 if len(entry) == 3:
                     # add a 4th item, `extra_filters`, to make it conform to the
-                    # typical shape of `model_map` a 3-item model map entry is
-                    # allowed as syntactic sugar
+                    # typical shape of `model_map`
+                    # a 3-item model map entry is allowed as syntactic sugar
                     self.model_map.append(entry + ({},))
                 if len(entry) == 4:
                     self.model_map.append(entry)
--- a/decorators/classes.py
+++ b/decorators/classes.py
@@ -5,6 +5,7 @@ from functools import wraps
 import rollbar
 
 # Django Imports
+from django.core.exceptions import ObjectDoesNotExist
 from django.http import HttpResponseNotFound
 from django.shortcuts import (
     get_object_or_404,
@@ -91,7 +92,7 @@ class restful_obj_seo_redirect(object):
         return wrapped
 
 
-class resolve_records_from_url(object):
+class resolve_records_from_restful_url(object):
     """Resolve Records from URL Patterns
 
     Decorator for resolving records from URL.
@@ -126,7 +127,12 @@ class resolve_records_from_url(object):
 
     def __init__(self, model_map, content_type='text/html'):
         self.content_type = content_type
-        self.model_map = model_map
+        self.model_map = []
+        for item in model_map:
+            if isinstance(item, tuple) and len(item) == 3:
+                self.model_map.append(item + ({},))
+            else:
+                self.model_map.append(item)
 
     def __call__(self, view_fn):
         @wraps(view_fn)
@@ -142,13 +148,13 @@ class resolve_records_from_url(object):
                 obj = None
                 model_map = self.model_map
 
-            for item in model_map:
-                model_or_relation = item[0]
-                field = item[1]
-                url_key = item[2]
-                extra_filter = len(item) > 3 and item[3] or {}
-
-                value = kwargs.get(url_key)
+            for (
+                model_or_relation,
+                field,
+                path_arg_name,
+                extra_filter,
+            ) in model_map:
+                value = kwargs.get(path_arg_name)
 
                 # Raises exception and breaks loop
                 key, obj = self._resolve_object(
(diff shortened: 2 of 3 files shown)


api/utils.py
```diff
@@ -32,6 +31,7 @@
 
 class HtkJSONEncoder(serializers.json.DjangoJSONEncoder):
     def default(self, obj):
+        from htk.models import HtkBaseModel
```
goztrk, line 34: # Reviewer Note
Moved import here because it was getting error

decorators/classes.py
```diff
@@ -69,21 +87,140 @@ def wrapped(*args, **kwargs):
                 kwargs[retrieved_key] = obj
                 response = view_fn(*args, **kwargs)
             return response
+
+        return wrapped
+
+
+class resolve_records_from_url(object):
```
goztrk, line 94: # Reviewer Note
Main Change

utils/text/transformers.py
```diff
@@ -180,4 +184,12 @@ def snake_case_to_lower_camel_case(string):
     """Convert `snake_case` string to `camelCase`"""
     camel_string = snake_case_to_camel_case(string)
     camel_string = camel_string[0].lower() + camel_string[1:]
-    return camel_string
\ No newline at end of file
+    return camel_string
+
+
+def pascal_case_to_snake_case(string):
```
goztrk, line 190: # Reviewer Note
Main change. Others are just to make Flake8 happy.

decorators/classes.py
```diff
@@ -69,21 +87,140 @@ def wrapped(*args, **kwargs):
...
+            if obj is None
+            else getattr(obj, model_or_relation)
+        )
+        DoesNotExist = (
+            model.model.DoesNotExist
+            if model.__class__.__name__ == 'RelatedManager'
+            else model_or_relation.DoesNotExist
+        )
+
+        filters = extra_filters.copy()
```
jontsai, line 190: Why copy?

decorators/classes.py
```diff
@@ -69,21 +87,140 @@ def wrapped(*args, **kwargs):
...
+
+    def _resolve_object(
+        self, obj, model_or_relation, field, value, extra_filters
+    ):
+        model = (
+            model_or_relation.objects
+            if obj is None
+            else getattr(obj, model_or_relation)
+        )
+        DoesNotExist = (
```
jontsai, line 184: Do we need to do this, or is there a Django class that is the base class of all of these?

utils/text/transformers.py
jontsai, line 190: Use the variable name `s` instead of `string`

jontsai (changes requested):

decorators/classes.py
```diff
@@ -69,21 +87,140 @@ def wrapped(*args, **kwargs):
...
+                model_map = self.model_map[1:]
+            else:
+                obj = None
+                model_map = self.model_map
+
+            for item in model_map:
+                model_or_relation = item[0]
+                field = item[1]
+                url_key = item[2]
+                extra_filter = len(item) > 3 and item[3] or {}
```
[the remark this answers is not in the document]
goztrk: The first 3 items in tuple is required and 4th one is optional.
There cannot be a 5th tuple item for now. Maybe in future developments.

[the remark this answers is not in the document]
goztrk: I added a mapper to `__init__` function to convert all of the tuples to 4 item tuples if `extra_filters` is not provided.

decorators/classes.py
```diff
@@ -69,21 +87,140 @@ def wrapped(*args, **kwargs):
...
+                obj = kwargs.get(self.model_map[0])
+                model_map = self.model_map[1:]
+            else:
+                obj = None
+                model_map = self.model_map
+
+            for item in model_map:
+                model_or_relation = item[0]
+                field = item[1]
+                url_key = item[2]
```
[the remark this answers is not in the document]
goztrk: > I added a mapper to `__init__` function to convert all of the tuples to 4 item tuples if `extra_filters` is not provided.

decorators/classes.py
```diff
@@ -69,21 +87,140 @@ def wrapped(*args, **kwargs):
...
+            # If first item is just a string, it means we need to get the object
+            # from kwargs
+            if isinstance(self.model_map[0], str):
+                obj = kwargs.get(self.model_map[0])
+                model_map = self.model_map[1:]
+            else:
+                obj = None
+                model_map = self.model_map
+
+            for item in model_map:
```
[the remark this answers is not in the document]
goztrk: My thought process was `extra_filters` would not always need to be provided. So the map would be needlessly crowded.

I added an empty dict in the `__init__` function if the provided tuple contains 3 items. I also added type guards to ensure all passed map items are expected type.

decorators/classes.py
goztrk: Since `extra_fields` is a `dict` we are doing operations on a reference and it might actually edit the original and mess up.

decorators/classes.py
goztrk: Done. Used `django.core.exceptions.ObjectDoesNotExist`.

utils/text/transformers.py
goztrk: Done

utils/text/transformers.py
```diff
@@ -168,16 +172,22 @@ def _repl(matchobj):
...
     """Convert `snake_case` string to `camelCase`"""
-    camel_string = snake_case_to_camel_case(string)
+    camel_string = snake_case_to_camel_case(s)
     camel_string = camel_string[0].lower() + camel_string[1:]
-    return camel_string
\ No newline at end of file
+    return camel_string
+
+
+def pascal_case_to_snake_case(s):
```
jontsai, line 188: Does this also work for `camel_case_to_snake_case`? Anyway, let’s add a wrapper and make whichever one is the more basic one the base implementation that gets wrapped, and the more complex one the wrapper on the outer layer.

jontsai (changes requested):

utils/text/transformers.py
```diff
@@ -168,16 +172,22 @@ def _repl(matchobj):
     return tokenized_title
 
 
-def snake_case_to_camel_case(string):
+def snake_case_to_camel_case(s):
     """Convert `snake_case` string to `CamelCase`"""
```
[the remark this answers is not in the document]
goztrk: I made those functions for JS URL exporter. I would like to fix it in a separate PR if that is ok?

utils/text/transformers.py
goztrk: No, it does not work unfortunately

decorators/classes.py
```diff
@@ -69,21 +89,161 @@ def wrapped(*args, **kwargs):
...
+
+    def __init__(self, model_map, content_type='text/html'):
+        self.content_type = content_type
+        self.model_map = []
+        for index, entry in enumerate(model_map):
+            if isinstance(entry, tuple):
+                if len(entry) == 3:
+                    # add a 4th item, `extra_filters`, to make it conform to the
+                    # typical shape of `model_map` a 3-item model map entry is
+                    # allowed as syntactic sugar
```
jontsai, line 137:
```suggestion
                    # typical shape of `model_map`
                    # a 3-item model map entry is allowed as syntactic sugar
```

decorators/classes.py
```diff
@@ -69,21 +89,161 @@ def wrapped(*args, **kwargs):
...
+            elif isinstance(entry, str):
+                if index == 0:
+                    self.model_map.append(entry)
+                else:
+                    raise ValueError(
+                        f'[{self._get_class_name()}] Only first item can be a string, got {entry!r}'
+                    )
+            else:
+                raise ValueError(
+                    f'[{self._get_class_name()}] Invalid model map item type: {type(entry)}'
```
jontsai, line 154:
```suggestion
                    f'[{self._get_class_name()}] Invalid model map entry type: {type(entry)}'
```

PR #404, opened 2023-11-30
fix: 🐛 Make StringIO imports python 2 and 3 compatible

goztrk:
## Status
**READY**

## Description
Fixed `cStringIO` imports to make it compatible with Python 2.x and Python 3.x

`cStringIO` module changed to `io.StringIO` in Python 3.

Used `six` package to import which is a compatibility layer for Python 2 and 3 apps.


jontsai (approved):
LGTM

PR #406, opened 2023-12-05
feat(Utils): ✨ Add New Function to Verify Account

lleung030:
### **Status**

Ready

### **Description**

- Added an account verification function


lib/gravatar/utils.py
```diff
@@ -24,7 +24,11 @@ def get_gravatar_hash(email):
     return gravatar_hash
 
 
-def get_gravatar_for_email(email, size=GRAVATAR_DEFAULT_SIZE):
+def get_gravatar_for_email(email, size=GRAVATAR_DEFAULT_SIZE, default=GRAVATAR_DEFAULT_IMAGE):
+    return build_gravatar_url_for_email(email, size=GRAVATAR_DEFAULT_SIZE, default=GRAVATAR_DEFAULT_IMAGE)
+
+
+def build_gravatar_url_for_email(email, size=GRAVATAR_DEFAULT_SIZE, default=GRAVATAR_DEFAULT_IMAGE):
```
jontsai, line 31: Line 47 below needs to be changed to `'default': default,` so that it can be overridden with the parameter passed in.

jontsai (changes requested):

PR #407, opened 2023-12-08
refactor: ♻️ Make b64encode Python 2/3 compat

goztrk:
## Status
**READY**

## Description
`base64.b64encode` requires encoded `bytes` in Python 3 but it works with `str` in Python 2.
The result of this function is `bytes` in Python 3 and `str` in Python 2.

Encoding the data with `utf-8` before passing to `base64.b64encode` works both for Python 2 and 3.
If the returning value is `bytes` (this only happens with Python 3) it needs to be decoded with `utf-8`

### DIFF
--- a/apps/async_task/utils.py
+++ b/apps/async_task/utils.py
@@ -8,21 +8,34 @@ def build_async_task_result(content, content_type, filename):
 
     This is necessary if we want to return multiple values, as the result by default is just a plain string.
     """
+    b64encoded_content = base64.b64encode(content.encode('utf-8'))
+
+    # Python 2 returns ``str`` and Python 3 returns ``bytes``
+    if isinstance(b64encoded_content, bytes):
+        b64encoded_content = b64encoded_content.decode('utf-8')
+
     payload = {
-        'content' : base64.b64encode(content),
-        'content_type' : content_type,
-        'filename' : filename,
+        'content': b64encoded_content,
+        'content_type': content_type,
+        'filename': filename,
     }
     result = json.dumps(payload)
     return result
 
+
 def extract_async_task_result_json_values(result_data):
-    """Companion function to perform the inverse of `build_async_task_result()`
-    """
+    """Companion function to perform the inverse of `build_async_task_result()`"""
     payload = json.loads(result_data)
 
-    content = base64.b64decode(payload['content'])
+    content = base64.b64decode(payload['content'].encode('utf-8'))
+    if isinstance(content, bytes):
+        content = content.decode('utf-8')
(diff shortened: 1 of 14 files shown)


lib/dynamic_screening_solutions/api.py
```diff
@@ -62,7 +62,7 @@ def _make_authorization_key(self, headers, secret_key):
         )
         signature = base64.b64encode(
             hmac.new(
-                bytes(secret_key), base_string, digestmod=hashlib.sha1
+                secret_key.encode('utf-8'), base_string.encode('utf-8'), digestmod=hashlib.sha1
```
jontsai, line 65:
```suggestion
                secret_key.encode(), base_string.encode(), digestmod=hashlib.sha1
```

lib/dynamic_screening_solutions/utils.py
```diff
@@ -26,8 +26,8 @@ def validate_webhook_request(request):
 
     signature = base64.b64encode(
         hmac.new(
-            bytes(hash_key),
-            request.body,
+            hash_key.encode('utf-8'),
+            request.body.encode('utf-8'),
```
jontsai, line 30:
```suggestion
            hash_key.encode(),
            request.body.encode(),
```

apps/async_task/utils.py
```diff
@@ -8,21 +8,34 @@ def build_async_task_result(content, content_type, filename):
 
     This is necessary if we want to return multiple values, as the result by default is just a plain string.
     """
+    b64encoded_content = base64.b64encode(content.encode('utf-8'))
+
+    # Python 2 returns ``str`` and Python 3 returns ``bytes``
```
jontsai, line 13:
```suggestion
    # Python 2 returns `str` and Python 3 returns `bytes`
```

lib/zillow/zillow_types.py
```diff
@@ -537,8 +546,12 @@ def exportSimple(self, outfile, level, name):
             outfile.write('<%s>%g</%s>' % (
                 self.name, self.value, self.name))
         elif self.content_type == MixedContainer.TypeBase64:
-            outfile.write('<%s>%s</%s>' % (
-                self.name, base64.b64encode(self.value), self.name))
+            value = b64encode(self.value)
+            if isinstance(value, bytes):
```
jontsai, line 550: Is this still needed if you're using `htk.compat.b64encode` ?

jontsai (changes requested):

jontsai (approved):
LGTM

PR #412, opened 2023-12-11
refactor: ♻️ AESCipher Python 2/3 compatability

goztrk:
## Status
**READY**

## Description
- Fixed `decrypt` method of `AESCipher` to make it compatible with Python 3.

It was decoding with `base64.b64decode` in Python 2. I updated it to `htk.compat.b64decode` but it needs to stay as bytes for `decrypt`ing to work.

**NOTE:** Tested with Python 2.7 and Python 3.9

### DIFF
--- a/compat.py
+++ b/compat.py
@@ -25,7 +25,7 @@ IS_PYTHON_2 = is_python2()
 IS_PYTHON_3 = is_python3()
 
 
-def b64encode(content, url_safe=False):
+def b64encode(content, url_safe=False, as_str=True):
     """Base64 encoder with Python 2/3 compatibility
 
     Args:
@@ -44,13 +44,14 @@ def b64encode(content, url_safe=False):
     )
     encoded = encoder(content)
 
-    # The result of Python 3 `base64.b64encode` is `bytes`, but we always want a `str`
-    encoded = encoded.decode() if IS_PYTHON_3 else encoded
+    # The result of Python 3 `base64.b64encode` is `bytes` but,
+    # most of the time we want a `str`.
+    encoded = encoded.decode() if IS_PYTHON_3 and as_str else encoded
 
     return encoded
 
 
-def b64decode(encoded, url_safe=False):
+def b64decode(encoded, url_safe=False, as_str=True):
     """Base64 decoder with Python 2/3 compability
 
     Args:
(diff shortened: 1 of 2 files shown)


utils/crypto.py
```diff
@@ -18,25 +16,33 @@ class AESCipher(object):
...
+        iv = Random.new().read(self.bs)
         cipher = AES.new(self.key, AES.MODE_CBC, iv)
-        return b64encode(iv + cipher.encrypt(raw.encode()))
+        encoded = b64encode(iv + cipher.encrypt(raw.encode()))
+        return encoded
 
     def decrypt(self, enc):
-        enc = b64decode(enc)
-        iv = enc[:AES.block_size]
+        # base64 encoded data is not convertible to `str` so not using `htk.compat.b64decode`
```
jontsai, line 33: Why? How so?

compat.py
```diff
@@ -44,13 +44,14 @@ def b64encode(content, url_safe=False):
     )
     encoded = encoder(content)
 
-    # The result of Python 3 `base64.b64encode` is `bytes`, but we always want a `str`
-    encoded = encoded.decode() if IS_PYTHON_3 else encoded
+    # The result of Python 3 `base64.b64encode` is `bytes` but,
+    # most of the time we want a `str`.
```
jontsai, line 48: Remove inline comment here, mention in docstring.

```
as_str: bool
    When `True` (default), converts the encoded result to a `str` for convenience.
    When `False`, the result is preseved as `bytes`

Returns: str or bytes
```

compat.py
```diff
@@ -64,7 +65,8 @@ def b64decode(encoded, url_safe=False):
     # `base64.b64decode` can decode either `str` or `bytes`, so no pre-treatment needed
     decoded = decoder(encoded)
 
-    # The result of Python 3 `base64.b64decode` is `bytes`, but we always want a `str`
-    decoded = decoded.decode() if IS_PYTHON_3 else decoded
+    # The result of Python 3 `base64.b64decode` is `bytes` but,
+    # most of the time we want a `str`.
```
jontsai, line 69: See above

jontsai (changes requested):

jontsai (approved):
LGTM

PR #413, opened 2023-12-21
feat: ✨ Add bytes decode guard to `seo_tokenize`

goztrk:
## Status
**READY**

## Description
- Added `decode` to the `title` parameter of `seo_tokenize` util.

Some functions are sending `bytes` to this function. Instead of trying to fix all possible places, added a guard to the `seo_tokenize` object.

### DIFF
--- a/utils/text/transformers.py
+++ b/utils/text/transformers.py
@@ -2,7 +2,10 @@
 import re
 
 # HTK Imports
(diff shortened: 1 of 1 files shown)


jontsai (changes requested):
IMO it should be handled at the call site. This function should strictly take a `str`

PR #417, opened 2024-01-19
fix: 🐛 datatables api view builder

goztrk:
## Description
- fixed base filter is being excluded instead of filtering.
- `model_class` is now can be a `QuerySet` to allow special circumstances like aggregation might be needed beforehand.


jontsai (approved):
LGTM

PR #424, opened 2024-04-10
✨ : updates `ForumMessage` model with foreign key `reply_to`

shreyastelkar:
### Description
- Adds a `reply_to` ForeignKey to ForumMessages to track messages that are replies to messages similar to replies on other forums such as `skool` and `reddit`. 
- Replies allow for forum users to engage in discussions under a message rather than individual messages under a thread. 
- Fixes code in the `organizations` abstract models so that the organization member can be correctly displayed when the `__str__` method is used. 
- Refer to https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.ForeignKey for more information on the usage of foreign keys in Django models.


jontsai (approved):
LGTM

PR #430, opened 2024-05-30
feat: ✨ adds `BaseConversationMessageReaction` to the `Conversation` models

shreyastelkar:
## Description:
- Adds a `BaseConversationMessageReaction` model to the `Conversation`s model that stores an `emoji shortcode` representation of an emoji reaction to a `Conversation Message`. 
- Uses the emoji library to emojize and demojize emojis for their shortcode and actual representation. This is added to the project requirements. 
- Adds `add_reaction` and `remove_reaction` model functions to add a reaction if it does not exist and delete a reaction if it exists.


apps/conversations/models.py
```diff
@@ -157,6 +163,31 @@ def was_edited(self):
...
+        This method is idempotent. If it is called multiple times with the same arguments, it will remove the record of the emoji reaction.
+
+        If it is called multiple times, and the reaction has already been deleted, this method has no effect and just performs a no-op.
+        """
+        reaction = self.emoji_reactions.filter(
+            user=user,
+            emoji_shortcode=emoji_shortcode,
+        ).first()
+        if reaction:
+            reaction.delete()
```
jontsai, line 189:
```suggestion
        reaction = self.emoji_reactions.filter(
            user=user,
            emoji_shortcode=emoji_shortcode,
        ).delete()
```

See: https://docs.djangoproject.com/en/5.0/topics/db/queries/#deleting-objects
> You can also delete objects in bulk. Every [QuerySet](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet) has a [delete()](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.delete) method, which deletes all members of that [QuerySet](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet).

Also: https://docs.djangoproject.com/en/5.0/ref/models/querysets/#delete
> The delete() method does a bulk delete and does not call any delete() methods on your models. It does, however, emit the [pre_delete](https://docs.djangoproject.com/en/5.0/ref/signals/#django.db.models.signals.pre_delete) and [post_delete](https://docs.djangoproject.com/en/5.0/ref/signals/#django.db.models.signals.post_delete) signals for all deleted objects (including cascaded deletions).

Depending on the implementation/needs, we might want to hook up with `post_delete` to update clients who have that emoji status.

apps/conversations/models.py
```diff
@@ -166,3 +197,29 @@ def save(self, **kwargs):
...
         self.conversation.save()
+
+
+class BaseConversationMessageReaction(models.Model):
+    """An emoji reaction to a message in
+    a conversation.
+    """
+
+    message = fk_conversation_message(
+        related_name='emoji_reactions', required=True
```
jontsai, line 208:
```suggestion
        related_name='reactions', required=True
```

apps/conversations/models.py
```diff
@@ -157,6 +163,31 @@ def was_edited(self):
     def is_deleted(self):
         return self.deleted_at is not None
 
+    def add_reaction(self, user, emoji_shortcode):
+        """Adds a reaction by `user` to this message.
+
+        This method is idempotent. If it is called multiple times with the same arguments, it will only create 1 record of the emoji reaction.
+        """
+
+        self.emoji_reactions.get_or_create(
```
jontsai, line 172:
```suggestion
        self.reactions.get_or_create(
```

apps/conversations/models.py
```diff
@@ -166,3 +197,29 @@ def save(self, **kwargs):
...
+
+    def __str__(self):
+        value = '%s' % emoji.emojize(self.emoji_shortcode)
+        return value
+
+    # TODO: add method to convert the emoji to shortcode
+    # @property
+    # def emoji_to_shortcode(self, emoji):
+    #     value = '%s' % emoji.demojize(emoji)
+    #     return value
```
jontsai, line 225:
```suggestion
    @property
    def emoji_to_shortcode(self, s):
        value = emoji.demojize(s)
        return value
```

Was this marked as `TODO` because of name conflicts? Simple fix -- just pick a different name for the parameter.

However, what is the purpose of this function existing? How would it ever get called? Seeing that it only takes in an external argument, and does not reference any attribute of this object (e.g. `self.*`) at all, it seems like something like this is more suitable as a utility function rather than an instance method.

jontsai (changes requested):

apps/conversations/models.py
shreyastelkar: Will look through the docs to understand things better.

apps/conversations/models.py
```diff
@@ -166,3 +196,21 @@ def save(self, **kwargs):
...
         self.conversation.save()
+
+
+class BaseConversationMessageReaction(models.Model):
+    """An emoji reaction to a message in
+    a conversation.
+    """
+
+    message = fk_conversation_message(related_name='reactions', required=True)
+    user = fk_user(related_name='reacted_by', required=False)
```
jontsai, line 207:
```suggestion
    user = fk_user(related_name='reactions', required=False)
```

This should be from the perspective of the object. `user.reactions` or `user.message_reactions` would be preferred, because `user.reacted_by` does not make sense.

jontsai (changes requested):

apps/conversations/models.py
shreyastelkar: Oh right, I should have been more careful of the foreign key names. I understand the `related_name` field and won't make this mistake again.

requirements.txt
```diff
@@ -13,3 +13,4 @@ requests>=2.31.0
 rollbar>=0.8.2
 stripe==1.19.0
 xhtml2pdf==0.0.6
+emoji>=2.12.1,<3
```
jontsai, line 16: keep this list sorted

jontsai (changes requested):

utils/text/unicode.py
```diff
@@ -1,11 +1,39 @@
...
+
+
+def demojize(s: str, shortest=True) -> str:
+    """Strips emojis from a string"""
+    demojized = emoji.demojize(s)
+    aliases = re.findall(r':\w+?:', demojized)
+    if aliases:
+        # Return the shortest alias
+        return min(aliases, key=len)
+    return demojized
```
jontsai, line 26:
```suggestion
    # return the shortest alias if they exist
    result = min(aliases, key=len) if aliases else demojized
    return result
```

jontsai (approved):
LGTM

PR #431, opened 2024-05-31
Pass AddEmail require_verification parameter via kwargs

YosefAshenafi:
## Status
Ready

## Description
Let we pass newly added parameter called `require_verification` via kwargs to sustain already existing code and usability.


apps/accounts/api/views.py
```diff
@@ -188,7 +188,7 @@ def unfollow(request, encrypted_uid):
 @login_required
 def email_add(request):
     user = request.user
-    add_email_form = AddEmailForm(user, request.POST)
+    add_email_form = AddEmailForm(user=user, data=request.POST)
```
jontsai, line 191: I think we should keep this the same because the positional arguments are likely being used by other projects as well.

Instead, I'd suggest changing https://github.com/hacktoolkit/django-htk/blob/c1262bf1e55a443d1fd80d58e35919c90ec33ad6/apps/accounts/forms/settings.py#L25C1-L33C43 to:

```
class AddEmailForm(forms.Form):
    email = forms.EmailField(label='Email')

    def __init__(self, user=None, *args, **kwargs):
        super(AddEmailForm, self).__init__(*args, **kwargs)
        self.user = user
        self.require_verification = kwargs.pop('require_verification', True)
        set_input_attrs(self)
        set_input_placeholder_labels(self)

    ...
```

At the call site, change it to `AddEmailForm(user, data, require_verification=<True|False>)`

jontsai (changes requested):

jontsai (approved):

jontsai (approved):

PR #432, opened 2024-06-11
feat: ✨ Make file blob contents editable

goztrk:
To make the files editable in the Django admin this small addition needed.
`editable` option is `False` by default for `BinaryField`


jontsai (approved):
LGTM

PR #434, opened 2024-06-12
features: 🌐 adds a template tag for localized strings (en-Us)

shreyastelkar:
- Adds a custom template tag to handle localized strings in templates.


templatetags/htk_tags.py
```diff
@@ -302,6 +302,23 @@ def get_request_duration():
     return duration
 
 
+@register.simple_tag()
+def localize(key):
+    '''TODO: handle this based on `by_language`
```
jontsai, line 307: Docstrings should always be surrounded by triple quotes

See https://peps.python.org/pep-0008/

templatetags/htk_tags.py
```diff
@@ -302,6 +302,23 @@ def get_request_duration():
     return duration
 
 
+@register.simple_tag()
+def localize(key):
```
jontsai, line 306: Please implement the TODO. You can change the the signature to `locale='en-US'` so that it is the default if not passed in.

templatetags/htk_tags.py
```diff
@@ -302,6 +302,23 @@ def get_request_duration():
...
+    '''TODO: handle this based on `by_language`
+    for any language
+    '''
+    from htk.apps.i18n.utils.data import retrieve_all_strings
+
+    language_code = 'en-US'
```
jontsai, line 312: rename to `locale`

templatetags/htk_tags.py
```diff
@@ -302,6 +302,23 @@ def get_request_duration():
...
+    localized_strings = retrieve_all_strings()
```
jontsai, line 313: Don't retrieve all strings just to extract 1; that's terribly inefficient.

Use the DB to look up just one record.

jontsai (changes requested):

templatetags/htk_tags.py
jontsai: Also, don't do the DB lookup directly here. Provide a utility function, and call that utility function here.

templatetags/htk_tags.py
```diff
@@ -302,6 +302,28 @@ def get_request_duration():
     return duration
 
 
+@register.simple_tag()
+def localize(key, locale='en-US'):
+
+    from htk.utils import resolve_model_dynamically
+    from django.core.exceptions import ObjectDoesNotExist
```
jontsai, line 309: This is not needed

templatetags/htk_tags.py
```diff
@@ -302,6 +302,28 @@ def get_request_duration():
...
+        htk_setting('HTK_LOCALIZED_STRING_MODEL')
+    )
+
+    try:
+        localized_string = LocalizedString.objects.get(
+            localizable_string__key=key,
+            language_code=locale,
+        ).value
+
+    except ObjectDoesNotExist:
```
jontsai, line 321:
```suggestion
    except LocalizedString.DoesNotExist:
```

The `ObjectDoesNotExist` exception is also available as a convenience accessor on the model.

jontsai (changes requested):

jontsai (changes requested):

jontsai (approved):
LGTM

PR #437, opened 2024-06-24
fix: 🔧 changes `early access` notification channel from `p4` to `p5` for dev and test servers

shreyastelkar:
## Description:
- Removes the explicit level `info` and uses the default level `debug` for the early access notification. 
- Default notification level: https://github.com/hacktoolkit/django-htk/blob/a3043a0dc859282f35827bb1dbd2cd81fd17eec1/utils/notifications.py#L37-L39

### DIFF
--- a/apps/prelaunch/models/base.py
+++ b/apps/prelaunch/models/base.py
@@ -169,7 +169,7 @@ class BasePrelaunchSignup(models.Model):
             'Early access has been enabled for {} <{}>'.format(
                 self.full_name, self.email
             ),
-            level='info',
+            level='debug',
         )
 
         try:
--- a/apps/prelaunch/models/base.py
+++ b/apps/prelaunch/models/base.py
@@ -15,7 +15,7 @@ from htk.apps.prelaunch.emails import (
     prelaunch_email,
 )
 from htk.utils import htk_setting
-from htk.utils.notifications import notify
+from htk.utils.notifications import slack_notify
 from htk.utils.request import get_current_request
 
 
@@ -139,8 +139,6 @@ class BasePrelaunchSignup(models.Model):
 
     def send_notifications(self):
         if htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
-            from htk.utils.notifications import slack_notify
-
             try:
                 slack_notify(self.notification_message)
             except:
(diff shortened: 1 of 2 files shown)


apps/prelaunch/models/base.py
```diff
@@ -169,7 +169,7 @@ def grant_early_access(self):
             'Early access has been enabled for {} <{}>'.format(
                 self.full_name, self.email
             ),
-            level='info',
+            level='debug',
```
jontsai, line 172: This changes it for every message. We only want to change the priority to lower if it is the dev environment.

One uses https://github.com/hacktoolkit/django-htk/blob/6c2d7c8a63253bf0e1f9398b72b42ba9dc8581f4/apps/prelaunch/models/base.py#L142

another uses https://github.com/hacktoolkit/django-htk/blob/6c2d7c8a63253bf0e1f9398b72b42ba9dc8581f4/apps/prelaunch/models/base.py#L168

We should probably just use one of them, and not both.

Something in this file is not working correctly: https://github.com/hacktoolkit/django-htk/blob/master/utils/notifications.py

jontsai (changes requested):

apps/prelaunch/models/base.py
```diff
@@ -139,8 +139,6 @@ def notification_message(self):
 
     def send_notifications(self):
         if htk_setting('HTK_SLACK_NOTIFICATIONS_ENABLED'):
-            from htk.utils.notifications import slack_notify
```
jontsai, line 142: If imports were done previously at the function level, I suggest leaving them there for now.

While it's generally better to import at the top of a file, there is a risk that it could have been done this way to avoid a circular dependency import. Or, it could have just been oversight. But we don't have enough context from just looking at this code.

utils/notifications.py
```diff
@@ -60,7 +60,12 @@ def slack_notify(message, level=None):
         default_level = (
             'debug' if (settings.ENV_DEV or settings.TEST) else 'info'
         )
-        level = level if level in channels else default_level
+        level = (
```
jontsai, line 63: I think this was fine from before?

apps/prelaunch/models/base.py
```diff
@@ -165,7 +163,7 @@ def grant_early_access(self):
         self.early_access = True
         self.early_access_code = uuid.uuid4().hex
         self.save()
-        notify(
+        slack_notify(
```
jontsai, line 166: I think the fix is to simply remove `level='info'` on line 170

jontsai (changes requested):

apps/prelaunch/models/base.py
shreyastelkar: Sounds good. Noted this to remember. Thanks.

jontsai (approved):

PR #438, opened 2024-06-25
fix: 🔧 modifies `as_dict` model method in the `reactions` model

shreyastelkar:
## Description
- Uses the `as_dict` method with a `user.profile` instead of `user` in the `ConversationReaction`s model.


jontsai (approved):
LGTM

PR #439, opened 2024-06-28
feat: ✨ adds additional email association` field to `create_test_user()`

shreyastelkar:
- Adds `should_associate_user_email` optional field to associate a `UserEmail` email to a test user created in `test_scaffold`.

### DIFF
--- a/test_scaffold/utils.py
+++ b/test_scaffold/utils.py
@@ -5,6 +5,7 @@ import random
 from django.contrib.auth import get_user_model
 
 # HTK Imports
+from htk.apps.accounts.utils import associate_user_email
 from htk.compat import uuid4_hex
 from htk.test_scaffold.test_data import *
 from htk.utils import htk_setting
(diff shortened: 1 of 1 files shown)


test_scaffold/utils.py
```diff
@@ -31,6 +33,10 @@ def create_test_user(
         email=email,
     )
 
+    if should_associate_user_email:
+        user_email = associate_user_email(user, email, confirmed=True)
+        user_email.set_primary_email()
```
jontsai, line 38: The `set_primary_email()` call is not needed.

This sets `user.email = user_email.email`; the former already has this value.

See: https://github.com/hacktoolkit/django-htk/blob/dc48e5f972805b1e0c6fe8f72e3b86b3fd862fbb/apps/accounts/models.py#L253-L272

jontsai (changes requested):

test_scaffold/utils.py
shreyastelkar: Got it. Thanks for providing this explanation, I see it now. I need to be careful when moving code over.

jontsai (approved):
LGTM

PR #441, opened 2024-07-02
feat: ✨ Adds a `BaseOrganizationJoinRequest` model to `Organizations`

shreyastelkar:
- Adds a join request model to handle requests made to join an organization. 
- Invitations and requests should cover the logistics to join an organization in multiple ways.


apps/organizations/models.py
```diff
@@ -214,7 +218,7 @@ class Meta:
 
     def __str__(self):
         value = (
-            '{organization_name} Member - {member_name} (member_email)'.format(
+            '{organization_name} Member - {member_name} {member_email}'.format(
```
jontsai, line 221: Why this change?

apps/organizations/models.py
```diff
@@ -335,6 +339,96 @@ def build_notification_message__declined(self):
...
+    email = models.EmailField(
+        blank=True, null=True, default=None
+    )  # the email where the request is sent
+    token = models.UUIDField(default=uuid.uuid4, unique=True)
+    accepted = models.BooleanField(
+        default=None, null=True
+    )  # True: accepted, False: declined, None: not responded yet
+    timestamp = models.DateTimeField(auto_now_add=True)
+    responded_at = models.DateTimeField(blank=True, null=True, default=None)
+    request_message = models.TextField(
```
jontsai, line 356: Not sure if this is needed.

jontsai (changes requested):

apps/organizations/models.py
shreyastelkar: Okay, will remove. I thought it could be cool to provide an optional introduction or bio when requesting to join.

apps/organizations/models.py
shreyastelkar: To correctly format the string with curly braces. I will wrap the braces with parentheses instead.

apps/organizations/models.py
shreyastelkar: I will just remove it.

apps/organizations/models.py
```diff
@@ -335,6 +337,87 @@ def build_notification_message__declined(self):
...
+        """Returns a dictionary that can be `json.dumps()`-ed as a JSON representation of this object"""
+        value = {
+            'id': self.id,
+            'organization': self.organization.name,
+            'user': self.user.profile.get_full_name() if self.user else None,
+            'email': self.email,
+            'accepted': self.accepted,
+            'requested_at': self.timestamp,
+            'responded_at': self.responded_at,
+            'request_message': self.request_message,
```
jontsai, line 373: Should we take this out?

I'm thinking of join requests sort of like "Applications"

e.g., a use case would be for a user to say, `I would like to "apply" to join your school, club, organization, etc.`

I think it makes sense to leave out the message so that subclasses can add additional fields to their forms if they would like, such as requiring the "applicant" to explain their experience and qualifications and why they should be considered for membership.

apps/organizations/models.py
```diff
@@ -335,6 +337,87 @@ def build_notification_message__declined(self):
...
+        )
+        return msg
+
+    def build_notification_message__created(self):
+        msg = self._build_notification_message(self.user, 'sent')
+        return msg
+
+    def build_notification_message__resent(self):
+        msg = self._build_notification_message(self.user, 're-sent')
+        return msg
```
jontsai, line 410: Not sure if the re-sent state makes sense.

It makes sense for invitations, that you can re-send, because the insider wants the outsider to join, and the invitation may have gone missing.

For join requests, where the outsider is seeking to gain entrance to become an insider, does it make sense to allow re-request? Probably not? If I applied 1M times to Stanford University, this would seem excessive and harassing their admissions office.

jontsai (changes requested):

apps/organizations/models.py
shreyastelkar: Right, that makes sense. What if there is an expiration for a request to join and then after that expiration time has ended, the user can make the request to join again. In this case, we would just use the normal request notification. I will remove this.

apps/organizations/models.py
shreyastelkar: Ah okay, will keep this in mind, thanks.

apps/organizations/models.py
```diff
@@ -335,6 +337,82 @@ def build_notification_message__declined(self):
...
+        )
+        return value
+
+    def json_encode(self) -> Dict[str, Any]:
+        """Returns a dictionary that can be `json.dumps()`-ed as a JSON representation of this object"""
+        value = {
+            'id': self.id,
+            'organization': self.organization.name,
+            'user': self.user.profile.get_full_name() if self.user else None,
+            'email': self.email,
```
jontsai, line 369: This field does not exist on `self`

Only include fields which exist.

apps/organizations/models.py
```diff
@@ -335,6 +337,82 @@ def build_notification_message__declined(self):
...
+
+    ##
+    # Notifications
+
+    def _build_notification_message(self, subject, action):
+        msg = '{subject_name} ({subject_username}<{email}>) request to join Organization <{organization_name}> has been {action}. Request Message - {message}'.format(  # noqa: E501
+            action=action,
+            subject_name=subject.profile.get_full_name(),
+            subject_username=subject.username,
+            email=subject.email,
```
jontsai, line 397: Does this still apply?

jontsai (changes requested):

apps/organizations/models.py
shreyastelkar: My bad.

apps/organizations/models.py
shreyastelkar: This would be the user's email which would still be applicable.

jontsai (approved):
LGTM

PR #444, opened 2024-07-28
feat: add ability to recognize early access code in header

goztrk:
## Description
Django cannot receive cookie data from the react-native environment.
Instead of adding it as a query-parameter on every request, sending it in the header was better.
This PR updates the `get_early_access_code` helper function to also get the early access code from header.


jontsai (approved):
LGTM

jontsai (approved):
LGTM

PR #446, opened 2024-08-06
feat: ✨ adds `remove_participant(s)` to the `Conversation` model

shreyastelkar:
- Adds model instance methods to remove a single participant or multiple participants from a conversation.


jontsai (approved):
LGTM

PR #451, opened 2024-09-30
feat: ✨ adds `ConversationParticipant fk_field` to `BaseConversation`s

shreyastelkar:
## Description
- Adds a foreign key for a `conversation participant` for any references to child models with the `conversation participant` foreign key `related name` field.


jontsai (approved):
LGTM

PR #455, opened 2024-10-24
Add field in Assessment Question model 

aarthi-axim:
Add field `is_enabled` in `AbstractAssessmentQuestion` model.


apps/assessments/models.py
```diff
@@ -55,6 +55,9 @@ class AbstractAssessmentQuestion(models.Model):
     # Optional "Exit message" to display when an incorrect answer is provided,
     # overrides `AbstractAssessment.knockout_message`
     knockout_message = models.TextField(blank=True, null=True, max_length=512)
+    # Must be true for a new question
+    # Set it to false when the question is answered and then modified
```
jontsai, line 59:
```suggestion
    # Must be `True` for a new question
    # Set it to `False` when disabling the question or the question is answered and then subsequently modified
```

jontsai (approved):

PR #460, opened 2025-01-15
feat ✨ Reset password mobile

YosefAshenafi:
## Status
Ready

## Description
Add a mobile deep link option on resetting password

### DIFF
--- a/apps/accounts/views.py
+++ b/apps/accounts/views.py
@@ -516,6 +516,7 @@ def reset_password(
     redirect_url_name='account_password_reset_success',
     template='account/reset_password.html',
     email_template=None,
+    mobile_reset_url_format=None,
     renderer=_r,
 ):
     """
@@ -538,6 +539,15 @@ def reset_password(
 
         if user:
             validlink = True
+
+            # Add mobile deep link URL if on mobile and format provided
+            if (request.user_agent.is_mobile or request.user_agent.is_tablet) and mobile_reset_url_format:
+                data['mobile_reset_url'] = mobile_reset_url_format.format(
+                    uid_base36=uidb36,
+                    token=token
+                )
+                data['appstore_url'] = htk_setting('HTK_APPSTORE_URL')
+
             if request.method == 'POST':
                 success, updated_user, form = reset_user_password(
                     request,
--- a/apps/accounts/views.py
+++ b/apps/accounts/views.py
@@ -540,11 +540,12 @@ def reset_password(
         if user:
             validlink = True
 
-            # Add mobile deep link URL if on mobile and format provided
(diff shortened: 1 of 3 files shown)


apps/accounts/views.py
```diff
@@ -538,6 +539,15 @@ def reset_password(
...
         if user:
             validlink = True
+
+            # Add mobile deep link URL if on mobile and format provided
+            if (request.user_agent.is_mobile or request.user_agent.is_tablet) and mobile_reset_url_format:
+                data['mobile_reset_url'] = mobile_reset_url_format.format(
+                    uid_base36=uidb36,
+                    token=token
+                )
+                data['appstore_url'] = htk_setting('HTK_APPSTORE_URL')
```
jontsai, line 549: Please add:

`htk/apps/mobile/constants/defaults.py`

The constant should have a default value of `None`

`HTK_APPSTORE_URL = None`

(New app under this dir: https://github.com/hacktoolkit/django-htk/tree/master/apps)

Then, include the import here:
 https://github.com/hacktoolkit/django-htk/blob/d231b9da22f2570ac81b08c49d359b56122523a2/constants/defaults.py#L104-L105

jontsai (changes requested):

apps/accounts/views.py
```diff
@@ -538,6 +539,15 @@ def reset_password(
 
         if user:
             validlink = True
+
+            # Add mobile deep link URL if on mobile and format provided
```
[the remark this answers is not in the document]
YosefAshenafi: The comment includes `if on mobile and format provided,"`which suits its placement. Changing it to "Add mobile deep link URL" only would make sense if moved below. Thoughts, @shreyastelkar?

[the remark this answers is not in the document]
jontsai: Place it inside the `if`

Inside the `if`, the comment is unambiguous as to what it applies to.

Outside of the `if`, it's hard to visually see if it covers only the immediate `if` block, or subsequent blocks at the same level of indentation as well.

Also, check throughout the codebase to see where the comments are typically placed. I have a feeling they are overwhelmingly inside.

apps/accounts/views.py
```diff
@@ -538,6 +539,15 @@ def reset_password(
...
+            if (request.user_agent.is_mobile or request.user_agent.is_tablet) and mobile_reset_url_format:
```
jontsai, line 544: Did this line run against Black?

jontsai (changes requested):

apps/accounts/views.py
YosefAshenafi: Applied now.

jontsai (approved):
LGTM

shreyastelkar (commented):
LGTM

PR #461, opened 2025-01-17
fix: 🔧 `profile user` reference in `is_organization_member`

shreyastelkar:
- Fixes a typo with the reference to the `profile user` when checking if the user is a member of an `organization`.


jontsai (approved):
LGTM

PR #462, opened 2025-01-21
feat: ✨ adds a custom channel parameter to `slack_notify`

shreyastelkar:
- Adds a new parameter to specify a custom Slack channel for sending the message, which takes precedence over other configured levels.

### DIFF
--- a/utils/notifications.py
+++ b/utils/notifications.py
@@ -48,10 +48,11 @@ def notify(message, level=None, use_messages=True, use_slack=None):
         slack_notify(message, level=level)
 
 
-def slack_notify(message, level=None):
+def slack_notify(message, level=None, custom_channel=None):
     """Send a Slack notification message
 
     `level` is one of ['critical', 'severe', 'danger', 'warning', 'info', 'debug',]
+    `custom_channel` is a string of a custom Slack channel name
     """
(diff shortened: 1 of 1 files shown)


utils/notifications.py
```diff
@@ -61,7 +62,7 @@ def slack_notify(message, level=None):
             'debug' if (settings.ENV_DEV or settings.TEST) else 'info'
         )
         level = level if level in channels else default_level
-        channel = channels.get(level, htk_setting('HTK_SLACK_DEBUG_CHANNEL'))
+        channel = custom_channel or channels.get(level, htk_setting('HTK_SLACK_DEBUG_CHANNEL'))
```
jontsai, line 65: FWIW there is a Python HTK Lite that has the interface that I want: https://github.com/hacktoolkit/python-htk/blob/master/htk/utils/slack.py

Unfortunately that would take maybe a day's worth of refactoring or more (to ensure existing integrations don't break); longer term, being able to call `send_message` (from: https://github.com/hacktoolkit/python-htk/blob/fe1934124fff169269f93b58d08af06a9871246b/htk/utils/slack.py#L139) would be more elegant.

jontsai (approved):
LGTM with comments.

PR #464, opened 2025-01-23
feat: ✨ adds a `mobile redirect url` to the `invitation response view` if the user is on mobile

shreyastelkar:
- Includes the mobile redirect URL in the `data` when the request to respond to an invitation comes from mobile.
- Adds an HTK setting for the `mobile url format` with it defaulted to the empty string.
- Refactors the `POST` endpoint code by using an instance method to `respond` to an invitation.


jontsai (approved):

apps/organizations/views.py
```diff
@@ -67,6 +67,13 @@ def setup(self, request, *args, **kwargs):
         self.data['invitation'] = self.invitation
 
     def get(self, request, *args, **kwargs):
+        if request.user_agent.is_mobile or request.user_agent.is_tablet:
+            token = self.invitation.token
+            self.data['mobile_invitation_response_url'] = (
+                self.mobile_invitation_response_url.format(token_hex=token.hex)
```
jontsai, line 73:
```suggestion
                self.mobile_invitation_response_url_format.format(token_hex=token.hex)
```

jontsai (changes requested):

apps/organizations/models.py
```diff
@@ -340,6 +340,27 @@ def status(self) -> str:
...
         return status
 
+    ##
+    # Invitation Response methods
+
+    def respond(self, accepted: str) -> None:
+        """Respond to an invitation with an accepted or declined status."""
+        request = get_current_request()
+        self.responded_at = timezone.datetime.now()
+        self.accepted = accepted == 'accept'
```
jontsai, line 350: - Passing a string to a model internal method is kind of an abstraction violation; should "peel the onion" before getting here
- Why is this method not on the `Invitation` model itself, but on the `Organization`?
- Getting the `user` from `request` is odd because... don't you always want the invitation user to be the user that responds?

apps/organizations/views.py
```diff
@@ -67,25 +65,27 @@ def setup(self, request, *args, **kwargs):
         self.data['invitation'] = self.invitation
 
     def get(self, request, *args, **kwargs):
+        is_mobile = request.user_agent.is_mobile or request.user_agent.is_tablet
+        mobile_url_format = htk_setting(
```
jontsai, line 69: dislike bc this makes the LHS/RHS names mismatch later

Maybe just go with boring (but consistent, symmetrical):

`mobile_invitation_response_url_format`

apps/organizations/views.py
```diff
@@ -67,25 +65,27 @@ def setup(self, request, *args, **kwargs):
...
-        self.invitation.user = request.user
-        if self.invitation.accepted:
-            # TODO: adding as member but maybe this should be a setting?
-            self.invitation.organization.add_member(
-                self.invitation.user, OrganizationMemberRoles.MEMBER
-            )
-
-        self.invitation.save()
-
+        self.invitation.respond(invitation_response)
```
jontsai, line 88: Add methods on the `Invitation` model:
- `accept`
- `decline`

jontsai (changes requested):

apps/organizations/models.py
shreyastelkar: - Right, makes sense. I initially had `accept` and `decline` methods but ended up creating a respond method for a one-liner call. 
- It is on the base invitation model since the HTK invitation response view uses the method as well.
- Yes, since there is validation with the `GET` endpoint that the user that was invited can only access the invitation, we can just set it to the invitation user. It was set as `request.user` before in the view to set the user that actually responded to the invitation, which should be the same as the invitation user.

jontsai (approved):
LGTM

PR #466, opened 2025-02-05
feat: ✨ adds a `RequestTooBig` Middleware to detect large request bodies

shreyastelkar:
- Adds `RequestTooBig` Middleware  to detect when the `request.body` triggers a `RequestDataTooBig` exception and returns a custom `JSON 400` response. 
- The location where the request.body validation occurs: https://github.com/django/django/blob/5f30fd2358fd60a514bdba31594bfc8122f30167/django/http/request.py#L368-L375

### DIFF
--- a/middleware/__init__.py
+++ b/middleware/__init__.py
@@ -2,6 +2,7 @@
 from htk.middleware.classes import (
     AllowedHostsMiddleware,
     GlobalRequestMiddleware,
+    RequestDataTooBigMiddleware,
     RequestTimerMiddleware,
     RewriteJsonResponseContentTypeMiddleware,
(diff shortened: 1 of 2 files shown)


jontsai (changes requested):
LFTM but you should just also insert it into `baws/settings.py` and test it.

- If a normal page loads fine, then you've tested a negative case.
- To test a positive case, follow the steps to use `curl` from here: https://chatgpt.com/share/67a2b598-ffb8-8005-ac59-86203c0759e6

jontsai (changes requested):

jontsai (approved):
LGTM

PR #470, opened 2025-02-27
fix: 🔧  related name for `organization members`

shreyastelkar:
- Fixes the incorrect `related_name` for the `user foreign key` on the `organization member`.


jontsai (approved):
LGTM

PR #473, opened 2025-03-18
feat: ✨ adds cm and mm conversion from meters

shreyastelkar:
- Adds the `cm` and `mm` properties on the `DistanceType` to conveniently convert a `DistanceType` instance in meters to one of the available units directly through a property.

### DIFF
--- a/utils/measurements/distance.py
+++ b/utils/measurements/distance.py
@@ -45,37 +45,37 @@ class DistanceType(Decimal):
     #
 
     @property
-    def m(self) -> Decimal:
-        return self
+    def m(self) -> 'DistanceType':
+        return self.from_meters(self)
 
     @property
-    def km(self) -> Decimal:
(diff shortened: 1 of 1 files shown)


utils/measurements/distance.py
```diff
@@ -52,6 +52,14 @@ def m(self) -> Decimal:
     def km(self) -> Decimal:
         return convert_unit(self, C.M_TO_KM)
 
+    @property
+    def cm(self) -> Decimal:
+        return convert_unit(self, C.M_TO_CM)
```
jontsai, line 57: I just had a thought; perhaps the return types of all of these should be `DistanceType`

```suggestion
    @property
    def cm(self) -> 'DistanceType:
        return self.__class__(convert_unit(self, C.M_TO_CM))
```

jontsai (changes requested):

jontsai (approved):
LGTM

PR #474, opened 2025-04-22
feat: ✨ refactors mapbox API response

shreyastelkar:
- Refactors the MapBox API response into its own function and returns the first accurate response in the `features` key. 
- Adds a method to return `context` data from the response in a dictionary for ease of access.
- https://docs.mapbox.com/api/search/geocoding

### DIFF
--- a/lib/mapbox/geocode.py
+++ b/lib/mapbox/geocode.py
@@ -172,6 +172,45 @@ def get_latlng(address, min_relevance_threshold=1):
 
 
 def fetch_mapbox_geocode_result(latitude, longitude):
+    """
+    Fetch the first mapbox geocode result for a given latitude and longitude
+
+    Example return value:
+        >>> fetch_mapbox_geocode_result(37.7749, -122.4194)
+        {'id': 'address.8939856834013852',
+         'type': 'Feature',
+         'place_type': ['address'],
+         'relevance': 1,
+         'properties': {'accuracy': 'point',
+         'mapbox_id': 'dXJuOm1ieGFkcjplNzRmN2U5MC1jODQzLTQxMWQtYTcyMi1mYTQwYTYwMzJjZGI'},
+         'text': 'Noriega St',
+         'place_name': '1818 Noriega St, San Francisco, California 94102, United States',
+         'center': [-122.41942, 37.774929],
+         'geometry': {'type': 'Point', 'coordinates': [-122.41942, 37.774929]},
+         'address': '1818',
+         'context': [{'id': 'neighborhood.601435372',
+         'mapbox_id': 'dXJuOm1ieHBsYzpJOWtzN0E',
+         'text': 'South of Market'},
+         {'id': 'postcode.8939856834013852', 'text': '94102'},
+         {'id': 'place.292358380',
+         'mapbox_id': 'dXJuOm1ieHBsYzpFVzBJN0E',
+         'wikidata': 'Q62',
+         'text': 'San Francisco'},
+         {'id': 'district.20547308',
+         'mapbox_id': 'dXJuOm1ieHBsYzpBVG1HN0E',
+         'wikidata': 'Q62',
+         'text': 'San Francisco County'},
+         {'id': 'region.419052',
+         'mapbox_id': 'dXJuOm1ieHBsYzpCbVRz',
+         'wikidata': 'Q99',
+         'short_code': 'US-CA',
+         'text': 'California'},
+         {'id': 'country.8940',
+         'mapbox_id': 'dXJuOm1ieHBsYzpJdXc',
+         'wikidata': 'Q30',
+         'short_code': 'us',
+         'text': 'United States'}]}
+    """
     extra_data = {
         'latitude': latitude,
         'longitude': longitude,
@@ -213,6 +252,13 @@ def fetch_mapbox_geocode_result(latitude, longitude):
 
 
 def reverse_geocode(latitude, longitude):
+    """
+    Reverse geocode a given latitude and longitude and return the address
+
+    Example return value:
+        >>> reverse_geocode(37.7749, -122.4194)
+        '1818 Noriega St, San Francisco, California 94102, United States'
+    """
     result = fetch_mapbox_geocode_result(latitude, longitude)
     if result:
         address = result.get('place_name')
--- a/lib/mapbox/geocode.py
+++ b/lib/mapbox/geocode.py
@@ -171,12 +171,11 @@ def get_latlng(address, min_relevance_threshold=1):
     return latlng
 
 
-def reverse_geocode(latitude, longitude):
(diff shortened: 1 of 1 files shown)


lib/mapbox/geocode.py
```diff
@@ -77,7 +77,7 @@ def main(argv=None):
         elif not is_geocode and len(args) == 2:
             latitude = args[0]
             longitude = args[1]
-            address = reverse_geocode(latitude, longitude)
+            address = reverse_geocode(latitude, longitude)['address']
```
jontsai, line 80: No, just return the whole dictionary/as much data as possible, normalized to what various API providers might return.

In practice, we might use several, not just MapBox.

lib/mapbox/geocode.py
```diff
@@ -171,6 +171,20 @@ def get_latlng(address, min_relevance_threshold=1):
     return latlng
 
 
+def extract_address_info(response):
```
jontsai, line 174: Does it make sense to get rid of information to go from more to less info?

Can we also include a link to the API docs in the docstring?

lib/mapbox/geocode.py
```diff
@@ -194,15 +208,14 @@ def reverse_geocode(latitude, longitude):
 
         results = response_json.get('features', [])
         if len(results) > 0:
-            result = results[0]
-            address = result['place_name']
+            info = extract_address_info(results)
```
jontsai, line 211: This breaks backwards compatibility for other code that is currently using this (unsure what); but in any case, if we are changing the shape of the return value, we should create a new function--unless we can change all callers.

I'd be curious to know what the immediate container with results is named.

lib/mapbox/geocode.py
```diff
@@ -194,15 +208,14 @@ def reverse_geocode(latitude, longitude):
 
         results = response_json.get('features', [])
         if len(results) > 0:
-            result = results[0]
```
jontsai, line 197: We go with `results[0]` because that is the most accurate/highest confidence result from the API.

lib/mapbox/geocode.py
jontsai: Also, naming inconsistency.

Function is called with `results` but parameter name is `response`.

lib/mapbox/geocode.py
```diff
@@ -171,6 +171,20 @@ def get_latlng(address, min_relevance_threshold=1):
...
 
 
+def extract_address_info(response):
+    """Extract address, city, and country_code from a Mapbox feature."""
+    result = {}
+    result['address'] = response[0].get('place_name')
+    for ctx in response[0].get('context', []):
+        ctx_id = ctx.get('id', '')
+        if ctx_id.startswith('place.'):
+            result['city'] = ctx.get('text')
```
jontsai, line 181: Assuming that it is a city is presumptuous. Different countries/places will have different hierarchies.

lib/mapbox/geocode.py
```diff
@@ -171,6 +171,20 @@ def get_latlng(address, min_relevance_threshold=1):
     return latlng
 
 
+def extract_address_info(response):
+    """Extract address, city, and country_code from a Mapbox feature."""
+    result = {}
+    result['address'] = response[0].get('place_name')
+    for ctx in response[0].get('context', []):
```
jontsai, line 178: Is this a list of dicts? Would be helpful to see API docs or sample response in the docstring

jontsai (changes requested):

lib/mapbox/geocode.py
```diff
@@ -171,12 +171,50 @@ def get_latlng(address, min_relevance_threshold=1):
...
+         'mapbox_id': 'dXJuOm1ieHBsYzpCbVRz',
+         'wikidata': 'Q99',
+         'short_code': 'US-CA',
+         'text': 'California'},
+         {'id': 'country.8940',
+         'mapbox_id': 'dXJuOm1ieHBsYzpJdXc',
+         'wikidata': 'Q30',
+         'short_code': 'us',
+         'text': 'United States'}]}
+    """
```
jontsai, line 213: Please add below `reverse_geocode`:

```
def reverse_geocode_with_context(latitude, longitude):
    geocode_result = fetch_mapbox_geocode_result(latitude, longitude)
    data = {}
    data['place_name'] = geocode_result['place_name']

    # Traverse through context and extract fields
    for field in geocode_result['context']:
        key = field['id'].split('.')[0]
        value = field['text']
        data[key] = value

        if 'short_code' in field:
            key = f'{key}_short'
            value = field['short_code']
            data[key] = value

    return data
```

Then you will have access to keys like `country, country_short, region, district`, etc

jontsai (changes requested):

lib/mapbox/geocode.py
```diff
@@ -171,12 +171,50 @@ def get_latlng(address, min_relevance_threshold=1):
     return latlng
 
 
-def reverse_geocode(latitude, longitude):
+def fetch_mapbox_geocode_result(latitude, longitude):
```
jontsai, line 174:
```suggestion
def fetch_mapbox_reverse_geocode_result(latitude, longitude):
```

lib/mapbox/geocode.py
```diff
@@ -171,12 +171,50 @@ def get_latlng(address, min_relevance_threshold=1):
...
+    """
+    Fetch the first mapbox geocode result for a given latitude and longitude
```
jontsai, line 176:
```suggestion
    Fetch the first Mapbox reverse geocode result for a given latitude and longitude
```

lib/mapbox/geocode.py
```diff
@@ -187,21 +225,45 @@ def reverse_geocode(latitude, longitude):
...
+
+def reverse_geocode(latitude, longitude):
+    """
+    Reverse geocode a given latitude and longitude and return the address
+
+    Example return value:
+        >>> reverse_geocode(37.7749, -122.4194)
+        '1818 Noriega St, San Francisco, California 94102, United States'
+    """
+    result = fetch_mapbox_geocode_result(latitude, longitude)
```
jontsai, line 262:
```suggestion
    result = fetch_mapbox_reverse_geocode_result(latitude, longitude)
```

jontsai (changes requested):

lib/mapbox/geocode.py
```diff
@@ -187,23 +226,81 @@ def reverse_geocode(latitude, longitude):
...
+
+    # Traverse through context and extract fields
+    for field in geocode_result['context']:
+        key = field['id'].split('.')[0]
+        value = field['text']
+        data[key] = value
+
+        if 'short_code' in field:
+            key = f'{key}_short'
+            value = field['short_code']
```
jontsai, line 299:
```suggestion
            value = field['short_code'].upper()
```

jontsai (approved):
LGTM

PR #475, opened 2025-05-01
fix: 🔧 uses `published_at` as the key when ordering `releases`

shreyastelkar:
- Removes the invalid ordering key called `version` which doesn't exist on the model and uses `published_at` instead. 
- Cleans up leftover code and imports.


YosefAshenafi (approved):
LGTM

jontsai (approved):

PR #477, opened 2025-05-15
fix 🔧  add user NoneType handling

jeffyang400:
Add handling to prevent a user NoneType assignment error from emerging when displaying or using the FitBitAPI

### DIFF
--- a/lib/fitbit/api.py
+++ b/lib/fitbit/api.py
@@ -32,7 +32,7 @@ class FitbitAPI(object):
         `client_id` OAuth2 Client Id from Fitbit App settings
         `client_secret` OAuth2 Client Secret from Fitbit App settings
         """
-        self.user = social_auth_user.user
+        self.user = social_auth_user.user if social_auth_user else None
         self.social_auth_user = social_auth_user
         self.client_id = client_id
         self.client_secret = client_secret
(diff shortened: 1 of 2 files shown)


lib/fitbit/api.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, social_auth_user, client_id, client_secret):
         `client_id` OAuth2 Client Id from Fitbit App settings
         `client_secret` OAuth2 Client Secret from Fitbit App settings
         """
-        self.user = social_auth_user.user
+        self.user = social_auth_user.user if social_auth_user else None
```
jontsai, line 35:
```suggestion
        if social_auth_user is None:
            raise ValueError("social_auth_user must be provided")

        self.user = social_auth_user.user
```

jontsai (changes requested):

jontsai (changes requested):
Actually, make this change here:

https://github.com/hacktoolkit/django-htk/blob/0c4020676aba7a8e0169fe4ebe3724f727678685/lib/fitbit/utils.py#L23

```
api = FitbitAPI(social_auth_user, client_id, client_secret) if social_auth_user else None

return api
```

jontsai (approved):
LGTM

PR #488, opened 2025-06-11
fix 🐛 Update import for InvitationStatus

YosefAshenafi:
- Update import for the InvitationStatus from apps invitation status instead of error throwing mesh.


jontsai (approved):
LGTM

PR #489, opened 2025-06-12
feat: Enhance InvitationsService to support multiple invitation models

baohouse:
- Added `init_models` method to `HtkBaseService` for dynamic model resolution.
- Updated `InvitationsService` to initialize models based on settings, allowing for multiple invitation models.
- Refactored email processing methods to iterate over available models for filtering invitations.


apps/invitations/services.py
```diff
@@ -9,8 +9,13 @@
 class InvitationsService(HtkBaseService):
     def __init__(self, *args, **kwargs):
         super(InvitationsService, self).__init__(*args, **kwargs)
+        model = htk_setting('HTK_INVITATION_MODEL')
+        models = htk_setting('HTK_INVITATION_MODELS')
+        if len(models) > 0:
+            self.init_models(models)
+        elif model:
+            self.init_models([model])
```
shreyastelkar, line 17:
```suggestion
            models = htk_setting('HTK_INVITATION_MODELS') or [htk_setting('HTK_INVITATION_MODEL')]
            if len(models) > 0:
                self.init_models(models)
```

shreyastelkar (approved):

services.py
```diff
@@ -12,3 +12,10 @@ def init_model(self, module_str):
         assert model is not None
 
         self.model = model
+
+    def init_models(self, module_strs):
```
jontsai, line 16: Refactor the single `init_model` to call the `init_models` method.

services.py
```diff
@@ -12,3 +12,10 @@ def init_model(self, module_str):
...
+        self.models = {}
```
jontsai, line 17: Interesting. I don't think we need to preserve the original mapping, because it can always be derived some other way. Just make it a list.

```
self.models = [resolve_model_dynamically(module_str) for module_str in module_strs]
assert all(self.models)
```

Will check that all are truthy (none are falsy, or `None`)

apps/invitations/services.py
```diff
@@ -60,7 +67,8 @@ def process_user_completed(self, user):
         """
         email = user.email
         if email:
-            q = self.model.objects.filter(email=email)
-            if q.exists():
-                invitation = q.first()
-                invitation.complete(user)
+            for model in self.models.values():
```
jontsai, line 70:
```suggestion
            for model in self.models:
```

Just go with a simple list, see below.

apps/invitations/services.py
```diff
@@ -19,24 +24,25 @@ def process_user_created(self, user):
         """
         email = user.email
         if email:
-            q = self.model.objects.filter(email=email)
-            if q.exists():
-                invitation = q.first()
+            for model in self.models.values():
```
jontsai, line 27:
```suggestion
            for model in self.models:
```

apps/invitations/services.py
jontsai: Yeah, check `if models and len(models) > 0:`

jontsai (changes requested):

jontsai (approved):
LGTM

PR #490, opened 2025-06-18
feat: Add `display` property to `HtkEnum` for symbolic name retrieval

baohouse:
This is especially helpful for displaying human-readable enums within Django templates. For example, let's say there's an object with a `status` enum property of a Model. So we define a `status_enum` on the model, and then within the Django template, we can use `model_instance.status_enum.display` to render the human-readable string.


jontsai (approved):
LGTM

PR #493, opened 2025-07-14
fix: 🔧  added cors headers and cache general function

jeffyang400:
Added CORS headers and a general cache function in HTK to clean up repeated code on both the HTK and Django sides.


jontsai (approved):
LGTM

PR #496, opened 2025-07-31
Adds Rollbar v3.0.0-alpha.4 JavaScript snippet

YosefAshenafi:
Adds Rollbar browser JS v3.0.0-alpha.4 snippet

Docs: http://rollbar.com/
Source: https://github.com/rollbar/rollbar.js/releases


baohouse:
I think this v3.0.0-alpha.4.html fragment has nearly no difference to the `v2.26.x.html` fragment. This is not really scalable. What you want is templates that relate to minimum configuration version, and then parse out the semver to use the minimum configuration version.

For example. let's say we rename the fragment to `min-2.26.html` meaning "at least 2.26". So if we pass in any of these versions "2.26.0", "2.26.4", "2.27", "3", or "3.0.0-alpha.4", they will map to `min-2.26.html`.

PR #497, opened 2025-08-01
feat: ✨ adds image metadata for facebook and twitter share images

shreyastelkar:


templates/fragments/meta.html
```diff
@@ -15,6 +15,14 @@
 {% if meta.image %}
 <meta property="og:image" content="{{ meta.image }}" />
 <meta name="twitter:image" content="{{ meta.image }}" />
+{# Facebook: 1.91:1 ratio (1200x628px) #}
+<meta property="og:image:width" content="1200" />
+<meta property="og:image:height" content="628" />
+{# Twitter: 16:9 ratio (1200x675px) #}
+<meta name="twitter:image:width" content="1200" />
+<meta name="twitter:image:height" content="675" />
```
baohouse, line 23: You're fixing the dimensions of the image, which could be different. For example if you upload a portrait or a panorama, it will not be these dimensions.

